Issue #6, opened 2022-03-09
Security review results


alexmurray:
As requested I have reviewed snapd-desktop-integration. I have done a basic code review, including running Coverity and other static analysis tools across the code, as well following the process outlined at https://wiki.ubuntu.com/SecurityTeam/Auditing (although as this isn't a debian package a lot of the tooling there is not relevant).

The main findings that I think need to be addressed are:

1. Source code files are missing license / copyright information
2. Possible use-after-free / double free of `notification` at [src/main.c:84](https://github.com/snapcore/snapd-desktop-integration/blob/main/src/main.c#L84)
  - if `action` is `yes` then a new `notification` will be allocated at line 56 which will shadow the one passed in as a function argument - instead I think either move the `g_object_unref()` from the end of this `notify_cb` function to the start OR use a different variable name to distinguish the `notification` as a function parameter and the one which gets created to tell the user the missing theme snaps are being installed
3. `check_themes_cb()` appears to leak the references to the various hash tables returned from `snapd_client_check_themes_finish()` - do these need to be unref'd before exiting from this function?
4. The use of libnotify as a dependency
  - Would it be better to just use `Gio.Notification` instead as I wonder if this is more maintained nowadays? I realise this would mean more work to make the whole application use `GApplication` but isn't this the future?

Finally, I believe this overall process needs to be driven by the MIR team so they may have other questions too. Standard security team MIR-like security review details as follows.

I reviewed snapd-desktop-integration at revision 7dcaa5e895f003494dc0ad8fead0c9f89ebade84.  This shouldn't be
considered a full audit but rather a quick gauge of maintainability.

snapd-desktop-integration is a small snap which watches the various desktop settings that specify what Gtk theme etc are chosen by the user. When this changes it queries snapd-glib to see if the theme which is chosen is currently installed and if not offers to install the themes as snaps if they are available, using notifications to interact with the user along the way.

- CVE History:
  - None
- Build-Depends
  - libnotify, snapd-glib
- pre/post inst/rm scripts
  - N/A - is a snap
- init scripts
  - N/A - is a snap
- systemd units
  - Yes (kind-of) - uses the snapd user-session daemon experimental feature
- dbus services
  - None
- setuid binaries
  - None (not allowed in snaps without an exception anyway)
- binaries in PATH
  - `/snap/bin/snapd-desktop-integration`
- sudo fragments
  - None
- polkit files
  - None
- udev rules
  - None
- unit tests / autopkgtests
[...]

PR #11, opened 2022-04-11
Don't pop up duplicate notifications

jhenstridge:
This PR cleans up the use of notifications, partly based on the feedback in #6.  In particular:

1. If the "Some required theme snaps are missing" notification is visible and we see another theme change, don't pop up a second notification. We use the `NotifyNotification::closed` signal to track when the notification has been closed, so a new notification can be presented if necessary.
2. Present the installation status notification as an update of the "Installing missing theme snaps" notification, so users don't end up with both notifications visible at once.

I haven't adopted the suggestion to use the GIO notifications API, since that does not currently work with snaps: it requires a D-Bus activatable desktop file, which is incompatible with the desktop file naming snapd currently uses. The xdg-desktop-portal notification portal doesn't change that requirement.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -37,8 +37,8 @@ ds_state_free(DsState *state)
     g_clear_pointer(&state->icon_theme_name, g_free);
     g_clear_pointer(&state->cursor_theme_name, g_free);
     g_clear_pointer(&state->sound_theme_name, g_free);
-    g_clear_pointer(&state->install_notification, g_object_unref);
-    g_clear_pointer(&state->progress_notification, g_object_unref);
+    g_clear_object(&state->install_notification);
+    g_clear_object(&state->progress_notification);
     g_free(state);
 }
 
@@ -59,13 +59,13 @@ install_themes_cb(GObject *object, GAsyncResult *result, gpointer user_data)
     }
 
     notify_notification_show(state->progress_notification, NULL);
-    g_clear_pointer(&state->progress_notification, g_object_unref);
+    g_clear_object(&state->progress_notification);
 }
 
 static void
 notification_closed_cb(NotifyNotification *notification, DsState *state) {
     /* Notification has been closed: */
-    g_clear_pointer(&state->install_notification, g_object_unref);
+    g_clear_object(&state->install_notification);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -41,18 +47,25 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC(DsState, ds_state_free);
...
     } else {
         g_print("Installation failed: %s\n", error->message);
-        notification = notify_notification_new("Installing missing theme snaps:", "Failed.", "dialog-information");
+        notify_notification_update(state->progress_notification, "Installing missing theme snaps:", "Failed.", "dialog-information");
     }
 
-    notify_notification_show(notification, NULL);
-    g_object_unref(notification);
+    notify_notification_show(state->progress_notification, NULL);
+    g_clear_pointer(&state->progress_notification, g_object_unref);
```
robert-ancell, line 62: Use `g_clear_object`.

src/main.c
```diff
@@ -41,18 +47,25 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC(DsState, ds_state_free);
...
+}
+
+static void
+notification_closed_cb(NotifyNotification *notification, DsState *state) {
+    /* Notification has been closed: */
+    g_clear_pointer(&state->install_notification, g_object_unref);
```
robert-ancell, line 68: Use `g_clear_object`.

src/main.c
```diff
@@ -87,9 +99,28 @@ notify_cb(NotifyNotification *notification, char *action, gpointer user_data) {
...
                                           (gchar**)icon_theme_names->pdata,
                                           (gchar**)sound_theme_names->pdata,
-                                          NULL, NULL, NULL, install_themes_cb, NULL);
+                                          NULL, NULL, NULL, install_themes_cb, state);
     }
-    g_object_unref(notification);
+}
+
+static void
+show_notification (DsState *state)
```
robert-ancell, line 107: Perhaps call `show_install_notification` - this isn't the only notification that is shown.

src/main.c
```diff
@@ -33,6 +37,8 @@ ds_state_free(DsState *state)
     g_clear_pointer(&state->icon_theme_name, g_free);
     g_clear_pointer(&state->cursor_theme_name, g_free);
     g_clear_pointer(&state->sound_theme_name, g_free);
+    g_clear_pointer(&state->install_notification, g_object_unref);
+    g_clear_pointer(&state->progress_notification, g_object_unref);
```
robert-ancell, line 41: `g_clear_object`

robert-ancell (changes requested):
Just some small stuff.

robert-ancell (approved):
LGTM

PR #17, opened 2022-09-23
Fix build with current snapd-glib version

sergio-costas:
The current versions of snapd-glib use, by default, libsoup3, which isn't available in Core20/Gnome-3-38. This fact breaks building of this snap.

This patch fixes this by forcing snapd-glib to use libsoup2.


kenvandine (approved):
LGTM

PR #18, opened 2022-09-23
DT-677 Migrate snapd-desktop-integration to Core22/Gtk4

sergio-costas:
This migration allows us to use libsoup3, which has support for UNIX sockets.

It also migrates it to Gtk4.

Also, this requires snapcraft 7.1.4, which is still in the "edge" channel.


kenvandine:
Don't we need to drop the libsoup-2 dep as well?

sergio-costas:
The tests are designed for libsoup-2, so unless we rewrite them, we still need libsoup-2.

snap/snapcraft.yaml
```diff
@@ -10,9 +10,16 @@ description: |
 grade: stable
 confinement: strict
 
+slots:
+  # for GtkApplication registration
+  snapd-desktop-integration:
+    interface: dbus
+    bus: session
+    name: io.snapcraft.desktop-integration
```
kenvandine, line 18: This should be io.snapcraft.SnapDesktopIntegration

snap/snapcraft.yaml
```diff
@@ -48,7 +53,9 @@ parts:
       - --prefix=/
     build-packages:
       - jq
+    build-environment:
+      - PKG_CONFIG_PATH: ${PKG_CONFIG_PATH:+$PKG_CONFIG_PATH:}$CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET/pkgconfig:$CRAFT_STAGE/usr/lib/pkgconfig
```
kenvandine, line 57: Shouldn't this be done already in the gnome extension?

kenvandine (changes requested):

kenvandine:
Please resolve the conflict as well

snap/snapcraft.yaml
sergio-costas: It should, but if I remove that...

imagen

I'll check the gnome extension, anyway.

sergio-costas: Yes. It was some garbage due to the change of the pkgconfig name for glib.

snap/snapcraft.yaml
sergio-costas: Absolutely right.

sergio-costas: Fixed.

snap/snapcraft.yaml
sergio-costas: Fixed

sergio-costas:
All fixed. Anyway, finally I'm using DBus instead of a HTTP Unix socket, so it would be possible to implement the functionality in the original GTK3 code. I'll discuss it with Michael Vogt and the others first, anyway.

hellsworth:
Looks good to me! Just need to rerun the job.

hellsworth:
again, seems fine but need to rerun the job. Perhaps for another PR, we should add a line in the github action yaml that allows us to manually kick off builds. To do this, add a `workflow_dispatch` line to test.yml like this:
```
diff --git a/.github/workflows/test.yml b/.github/workflows/test.yml
index b3a6ed4..c164384 100644
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
 
 jobs:
   snap:
```

sergio-costas:
I think that will be better to close this MR and https://github.com/snapcore/snapd-desktop-integration/pull/20 , and go "step by step" with MRs like https://github.com/snapcore/snapd-desktop-integration/pull/28 , where the code is migrated first in Gtk3 to the new APIs, and only then, in a much smaller step, can be migrated to Gtk4.

PR #20, opened 2022-09-28
Dt 576 add support for busy popups (core22)

sergio-costas:
This patch adds a DBus interface to allow to show a Gtk popup window when a Snap is being refreshed and can't be launched yet.

TODO:
* allow to show an icon
* don't show repeated windows
* group several programs in a single window


src/dbus.c
```diff
@@ -0,0 +1,89 @@
...
+register_dbus (GDBusConnection  *connection,
+               DsState          *state,
+               GError          **error)
+{
+    static const gchar introspection_xml[] =
+    "<node>"
+    "    <!-- io.snapcraft.SnapDesktopIntegration:"
+    "        @short_description: Interface for desktop integration of Snapd"
+    ""
+    "        This D-Bus interface allows to achieve better integration of"
```
mardy, line 36: Unless I'm terribly mistaken, these will be concatenated as "...integration of SnapdThis D-Bus interface..." unless you explicitly write a `\n` at the end of each line.

sergio-costas: Ops... you are right!

sergio-costas: Anyway, being a description, I don't think it is problematic: it is there just as "documentation", but the XML is directly consumed by GLib.

PR #24, opened 2022-10-06
DT-708: add gettext support

sergio-costas:
This MR adds support for internationalization/localization to snap-desktop-integration. It is a must, now that there are texts visible to the user.


sergio-costas:
@kenvandine Can you review this when you have some spare time?

kenvandine (approved):

PR #25, opened 2022-10-07
Don't restart if DISPLAY is not set

sergio-costas:
In some cases, like when connecting from a remote terminal using SSH, GTK can't be initialized and fails. In that case, this daemon keeps being relaunched over and over.

This patch fixes it by checking first if it is possible to initialize GTK, and only relaunching the daemon if there was an error.

Fix https://github.com/snapcore/snapd-desktop-integration/issues/14

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -21,7 +21,7 @@ apps:
     daemon: simple
     passthrough: #///! TODO: Remove once daemon-scope lands in snapcraft
       daemon-scope: user
-    restart-condition: always
+    restart-condition: on-failure
     plugs:
       - snap-themes-control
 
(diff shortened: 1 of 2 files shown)


sergio-costas:
The only problem that I see is that manually killing the process doesn't relaunch it...

jhenstridge (commented):
This looks fine. The only thing I think we'd need to be careful of is to make sure the daemon actually starts on session boot though. That is, to handle the following case:

1. systemd user session starts
2. snapd-desktop-integration starts
3. gnome-shell starts, and pushes DISPLAY/WAYLAND_DISPLAY to user session environment.

In this case, gtk_init_check() might return false, and we never try to restart. At present, snapd's user daemons code doesn't give us enough control to ensure (2) happens after (3), so I kind of wonder if we were relying on the busy loop restart-condition to actually start up.

PR #26, opened 2022-10-07
DT-576: don't duplicate refresh popups

sergio-costas:
This MR keeps track of the snaps that have a refresh popup, to avoid showing more than one.


sergio-costas:
@kenvandine and also review this one too...

kenvandine (approved):

PR #27, opened 2022-10-10
DT-576: Add close API

sergio-costas:
This MR adds a new DBus call that allows to close a "XXX is being refreshed" popup for a specific snap.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -20,6 +20,11 @@ handle_notifications_method_call(GDBusConnection       *connection,
         g_dbus_method_invocation_return_value(invocation, NULL);
         return;
     }
+    if (g_strcmp0 (method_name, "ApplicationRefreshCompleted") == 0) {
+        handle_close_application_window(parameters, (DsState*) user_data);
+        g_dbus_method_invocation_return_value(invocation, NULL);
+        return;
+    }
     g_dbus_method_invocation_return_error(invocation, G_DBUS_ERROR, G_DBUS_ERROR_UNKNOWN_METHOD, "Unknown method");
 }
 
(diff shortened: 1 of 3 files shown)


sergio-costas:
@kenvandine Can you review this when you have some spare time?

kenvandine:
@jhenstridge can you please review this?

src/dbus.c
```diff
@@ -59,7 +65,17 @@ register_dbus (GDBusConnection  *connection,
...
+    "            ApplicationRefreshCompleted:"
+    "            @application_name: the name of the application name"
+    "           @extra_parameters: a dictionary with extra optional parameters."
+    "           Method used to notify the daemon that the refresh is completed"
+    "           and the dialog can be closed."
+    "       -->"
+    "       <method name=\"ApplicationRefreshCompleted\">"
+    "       <arg direction=\"in\" type=\"s\" name=\"application_name\"/>"
+    "       <arg direction=\"in\" type=\"a{sv}\" name=\"extra_parameters\"/>"
     "        </method>"
```
jhenstridge, line 79: Rather than embedding the D-Bus introspection data like this, it'd be worth moving it to its own file and generating skeletons to plug the implementation into. This can be done using meson's gnome.gdbus_codegen() helper:

https://mesonbuild.com/Gnome-module.html#gnomegdbus_codegen

The skeleton code will then handle method dispatch, and we'd have a little type safety ensuring the method implementations match the signatures.

That's probably worth doing as a separate PR though, given that it is a cleanup for what's already in place.

jhenstridge (approved):
Looks good. I think there is room to clean up the details of how the D-Bus code is hooked up, but that can be dealt with later.

src/dbus.c
sergio-costas: Good idea!

PR #28, opened 2022-10-10
DT-719: Migrate into GtkApplication

sergio-costas:
Currently, snapd-desktop-integration uses an odd mix of glib mainloop and pieces of GTK. This must be updated to use GtkApplication, in order to do a future full migration into Gtk4.


sergio-costas:
@kenvandine Can you review this one when you have some spare time, please?

kenvandine (approved):
Thanks, looks good

PR #30, opened 2022-10-13
DT-731: fix segmentation fault

sergio-costas:
There seems to be a problem in g_hash_table_lookup when passing NULL as the key. According to the documentation, it is allowed, but here it crashes.

This is a quick patch for it, but a patch for GLib must be sent too.

Fix https://github.com/snapcore/snapd-desktop-integration/issues/15


kenvandine (approved):
Thanks!

PR #31, opened 2022-10-14
Always restart with a delay

sergio-costas:
This MR changes the way the daemon responds when it is killed: it will always restart, but with a 5-second delay to avoid consuming too many resources. This allows to guarantee that it is always launched, no matter if the user begins with a ssh session and, in the middle, changes to a graphical one.


sergio-costas:
@jhenstridge After what we talked the other day, I think that this can be a solution... What do you think?

kenvandine:
I think this is sensible, but what about a case where someone wants to stop the service manually? If stopped manually, will it automatically restart in this scenario?  If so, does that mean a user would need to disable the service to prevent restart?

sergio-costas:
@kenvandine  @jhenstridge  Ok, I found a case where the previous code doesn't work, and requires to merge this MR to fix it (and maybe with an smaller timeout): if the daemon/service is launched too soon in the session, gtk_init_check() will fail, the process will exit with a 0 return code, and systemd won't relaunch it, no matter if the session is a local graphical one.

kenvandine (approved):
LGTM

PR #32, opened 2022-10-17
DT-737: Migration to GDBus-codegen

sergio-costas:
Migrate the DBus code to GDBus-codegen.


src/dbus.c
```diff
@@ -17,106 +17,48 @@
...
-                                          node_info->interfaces[0],
-                                          &notifications_vtable,
-                                          (gpointer) state,
-                                          NULL,
-                                          error) == 0) {
-        g_prefix_error(error, "Failed to register notifications object: ");
-        return FALSE;
-   }
-   return TRUE;
+    SnapDesktopIntegration *skeleton = snap_desktop_integration_skeleton_new();
```
jhenstridge, line 53: It looks like this object is leaked. Perhaps return the object, store it in DsState, and unref it in ds_state_free()?

src/dbus.c
```diff
@@ -17,106 +17,48 @@
...
+    g_signal_connect(skeleton, "handle_application_is_being_refreshed",
+                     G_CALLBACK (dbus_handle_application_is_being_refreshed),
+                     state);
+    g_signal_connect(skeleton, "handle_application_refresh_completed",
+                     G_CALLBACK (dbus_handle_close_application_window),
+                     state);
+    g_dbus_interface_skeleton_export (G_DBUS_INTERFACE_SKELETON(skeleton),
+                                      connection,
+                                      "/io/snapcraft/SnapDesktopIntegration",
+                                      NULL);
```
jhenstridge, line 63: This call can fail in the same way that g_dbus_connection_register_object could. Should the error be checked/propagated?

jhenstridge (changes requested):
This looks like a good improvement. I've left a few comments below.

src/dbus.c
sergio-costas: Not really. This object has to live from the beginning to the end of the program, because it is the one that manages the DBus connections. DsState, instead, is the state of each progress window: one is created every time a progress window is shown, and destroyed when it is closed.

Of course, I can store the skeleton somewhere to unref() it when the program ends...

jhenstridge: There's just one `DsState` created in `main` and freed when main exits. That seems like the appropriate lifetime for the skeleton. Are you thinking of RefreshState here?

sergio-costas: Oh... you are right...

src/dbus.c
sergio-costas: Yes. Fixed.

jhenstridge (approved):
Looks good!

sergio-costas:
@jhenstridge Thanks!

PR #36, opened 2022-10-25
DT-774: install required theme snap on notification click

sergio-costas:
When a gtk-theme snap is missing, snapd-desktop-integration shows a notification allowing to install it. Unfortunately, when that notification is moved to the list of "old" notifications, it stops reacting.

The reason is that when a notification in that list is clicked, the program that created it receives a 'default' action, while the two actions defined in the notification are 'yes' and 'no'.

This patch adds a 'default' action that also installs the snap.

Fix https://github.com/snapcore/snapd-desktop-integration/issues/13

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -82,7 +82,7 @@ static void
 notify_cb(NotifyNotification *notification, char *action, gpointer user_data) {
     DsState *state = user_data;
 
-    if (strcmp(action, "yes") == 0) {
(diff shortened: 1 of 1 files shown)


hellsworth (approved):
lgtm!

kenvandine:
@sergio-costas the last merge caused a conflict, can you please rebase and fix the conflict?

PR #37, opened 2022-10-25
DT-775: show a meaningful message on error

sergio-costas:
If a gtk-theme snap is missing and snapd-desktop-integration offers the user to install it, the user clicks on "yes" and then it cancels the process (for example, when asked by the root password), the error message is meaninless: it just says "failed".

This patch fixes this, and puts the basic infraestructure to show more specific errors on other cases.

Fix https://github.com/snapcore/snapd-desktop-integration/issues/13


hellsworth (approved):
lgtm!

kenvandine:
@sergio-costas the last merge caused a conflict, can you please rebase and fix the conflict?

kenvandine (approved):
LGTM

PR #38, opened 2022-10-25
DT-776: Added missing translations

sergio-costas:
Some texts weren't translated.


sergio-costas:
@kenvandine Can you review this when you have some spare time, please?

kenvandine (approved):
Awesome, thanks!

PR #42, opened 2022-12-14
DT-906: avoid daemon being relaunched over and over

sergio-costas:
When the user enters a non-graphical session (like a SSH session), the GTK initialization in the daemon fails, and it exits with a "zero" error to ensure that systemd relaunches it after two seconds. The result is that, in those cases, the journal log is full of messages saying that snapd-desktop- integration has failed and is being reloaded, one every two seconds.

Also, since the systemd daemons are launched once per user, instead of once per session, we must keep reloading it even if the initialization fails, because it can happen that the user, while having the SSH session opened, enters into a desktop session. Not reloading it would mean that snapd- desktop-integration wouldn't be loaded in that case.

Also, sometimes the daemon is launched "too quickly" by systemd, and although it is a desktop session, the desktop isn't ready and gkt_init_check() fails. In that case we must force a reload to ensure that it is launched again when the desktop is ready.

This patch fixes it by monitoring the logind DBus interface to detect when a new session is opened.

It first calls gtk_init_check(). If it suceeds, then the daemon is running inside a desktop session and it can work as expected.

But if it fails, it can mean that we are inside a SSH session, or in a desktop session but the daemon was launched "too quickly". So in this case, the daemon tries to connect to the login1 daemon, which manages the sessions, and checks all the current sessions for the current user.

If any of the sessions is a graphical session (which is known by its type being "x11", "wayland" or "mir"), then the daemon knows that it must exit in order to be relaunched two seconds after, when the desktop is fully ready.

Instead, if there is no graphical session opened for the current user, the daemon concludes that it is running in a SSH, or in a local, text mode TTY, so it will connect to the "session-new" signal of the logind daemon and wait for new sessions being opened. When that happens, it will check if that new session is a desktop one. If it ins't (because, for example, the user opened another SSH connection), the daemon will just continue waiting for a new session. But if the new session is a desktop one, then it will exit to ensure that systemd will relaunch it.

It is a must to exit and let systemd to relaunch it because the environment variables are different when running in a SSH/TTY session than in a desktop session.

Finally, if the daemon is unable to connect/read data from the logind daemon, it will just exit, thus working like the current code. This can happen if the snap isn't connected to the "login-session-observe" interface.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -25,6 +25,7 @@ apps:
     restart-delay: 2s
     plugs:
       - snap-themes-control
+      - login-session-observe
 
 slots:
   snapd-desktop-integration:
--- a/data/meson.build
+++ b/data/meson.build
@@ -1,0 +1,12 @@
+gnome = import('gnome')
+
+# The returned source would be passed to another target
+login_src = gnome.gdbus_codegen('org.freedesktop.login1',
+  sources: 'org.freedesktop.login1.dbus.xml',
+  interface_prefix : 'org.freedesktop'
+)
+
+login_session_src = gnome.gdbus_codegen('org.freedesktop.login1.Session',
+  sources: 'org.freedesktop.login1.session.dbus.xml',
+  interface_prefix : 'org.freedesktop.login1.session'
+)
--- a/data/org.freedesktop.login1.dbus.xml
+++ b/data/org.freedesktop.login1.dbus.xml
@@ -1,0 +1,417 @@
+<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
+"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
+<node>
+ <interface name="org.freedesktop.DBus.Peer">
+  <method name="Ping"/>
+  <method name="GetMachineId">
+   <arg type="s" name="machine_uuid" direction="out"/>
+  </method>
+ </interface>
+ <interface name="org.freedesktop.DBus.Introspectable">
+  <method name="Introspect">
+   <arg name="xml_data" type="s" direction="out"/>
+  </method>
+ </interface>
+ <interface name="org.freedesktop.DBus.Properties">
+  <method name="Get">
+   <arg name="interface_name" direction="in" type="s"/>
+   <arg name="property_name" direction="in" type="s"/>
+   <arg name="value" direction="out" type="v"/>
+  </method>
+  <method name="GetAll">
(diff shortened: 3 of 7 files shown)


snap/snapcraft.yaml
```diff
@@ -25,6 +25,7 @@ apps:
     restart-delay: 2s
     plugs:
       - snap-themes-control
+      - login-session-observe
```
kenvandine, line 28: We're going to need to request autoconnect of this interface, which is fine.

sergio-costas: Anyway, the code is safe and still works if it isn't connected (although it will spit errors each two seconds, just like now).

kenvandine (approved):
LGTM

hellsworth:
In a Jammy vm, I see that if I am tracking the snapd-desktop-integration in stable (latest/stable/ubuntu-22.04 is the default), then I see the following issue:

* Reboot the VM and do not login via the login screen that would initiate the GDM session
* ssh to the system
* see no snapd-desktop-integration process
* watch the journalctl logs and see many of these:
```
Feb 21 11:06:20 jammy snapd-desktop-i[4048]: Not loading module "atk-bridge": The functionality is provided by GTK natively. Please try to not load it.
Feb 21 11:06:20 jammy snapd-desktop-integration.snapd-desktop-integration[4048]: Failed to do gtk init
Feb 21 11:06:22 jammy systemd[1372]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Scheduled restart job, restart counter is at 11.
Feb 21 11:06:22 jammy systemd[1372]: Stopped Service for snap application snapd-desktop-integration.snapd-desktop-integration.
Feb 21 11:06:22 jammy systemd[1372]: Started Service for snap application snapd-desktop-integration.snapd-desktop-integration.
Feb 21 11:06:22 jammy snapd-desktop-i[4113]: Not loading module "atk-bridge": The functionality is provided by GTK natively. Please try to not load it.
Feb 21 11:06:22 jammy snapd-desktop-integration.snapd-desktop-integration[4113]: Failed to do gtk init
Feb 21 11:06:24 jammy systemd[1372]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Scheduled restart job, restart counter is at 12.
Feb 21 11:06:24 jammy systemd[1372]: Stopped Service for snap application snapd-desktop-integration.snapd-desktop-integration.
Feb 21 11:06:24 jammy systemd[1372]: Started Service for snap application snapd-desktop-integration.snapd-desktop-integration.
```

Then if I refresh to the snapd-desktop-integration in candidate, reboot, and only ssh, I no longer see the error messages and I see the expected two snapd-desktop-integration processes.

Great work @sergio-costas !

PR #43, opened 2022-12-15
Migrate to Core22

sergio-costas:


snap/snapcraft.yaml
```diff
@@ -48,8 +48,6 @@ parts:
       - -Dsoup2=true
     build-packages:
       - libpolkit-gobject-1-dev
-    prime:
-      - usr/lib/*/libsnapd-glib.so.*
```
kenvandine, line 52: Do we need to stage more than just the lib here?

sergio-costas: You are right. Fixed.

PR #44, opened 2023-01-09
DT-954: add percentage support

sergio-costas:
This MR adds two new DBus methods that allow to set a percentage value in the progress bar, or set again the pulsing bar.

This will be useful to give to the user more info about how the snap download is going.

### DIFF
--- a/data/io.snapcraft.SnapDesktopIntegration.dbus.xml
+++ b/data/io.snapcraft.SnapDesktopIntegration.dbus.xml
@@ -41,5 +41,28 @@
             <arg direction="in" type="s" name="application_name"/>
             <arg direction="in" type="a{sv}" name="extra_parameters"/>
         </method>
+        <!--"
+             ApplicationRefreshPulsed:"
+             @application_name: the name of the application name"
+             @extra_parameters: a dictionary with extra optional parameters."
+             Method used to notify the daemon that the progress bar must be"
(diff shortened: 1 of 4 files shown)


hellsworth:
I've built and installed this snapd-desktop-integration. Note that to install it I needed to set experimental.user-daemons=true.

```
$ sudo snap set system experimental.user-daemons=true
$ sudo snap install snapd-desktop-integration_0.1_amd64.snap --dangerous
```

Then using `d-feet`, under the "Session Bus", I was able to send strings to a few different key components:
image

* ApplicationRefreshCompleted: `"firefox", "", {}` and we see the pop-up showing a downloading string with the bouncing blue bar
image

* Without closing the pop-up window, open a window for ApplicationRefreshPercentage: `"firefox", "downloading", 0.3, {}`. After clicking Execute, we see the bar update to reflect the correct percentage (at least visually). I tried this with a few different values and they all looked good.
image

* Again, without closing the pop-up window, open a window for ApplicationRefreshPulsed: `"firefox", "pulsing now", {}`. We see the blue bar start bouncing back and forth again and the string has been updated.
image

To close the pop-up, I opened a window again for ApplicationRefreshCompleted: `"firefox", {}` and Execute makes the window close as expected.

kenvandine (approved):

PR #47, opened 2023-03-03
Set size in table

sergio-costas:
Setting a size in the window is not enough to force a size. This MR fixes it.


robert-ancell (approved):
LGTM

PR #49, opened 2023-03-16
Add extra options

sergio-costas:
Allows to specify several extra options in the "extra_data" field in each DBus call. These options allow to set an icon, change the window's title, or the message.

image

In order to get access to any SNAP icon, it requires a "system-files" interface and permissions, which, for the snap store, requires specific permissions https://snapcraft.io/docs/system-files-interface

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -14,12 +14,6 @@ layout:
   /usr/share/locale:
     bind: $SNAP/usr/share/locale
 
-plugs:
-  read-snap-data:
-    interface: system-files
-    read:
-      - /snap
-
 apps:
   snapd-desktop-integration:
     extensions: [gnome]
@@ -32,7 +26,7 @@ apps:
     plugs:
       - snap-themes-control
       - login-session-observe
-      - read-snap-data
+      - desktop-launch
 
 slots:
   snapd-desktop-integration:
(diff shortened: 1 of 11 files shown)


snap/snapcraft.yaml
```diff
@@ -14,6 +14,12 @@ layout:
   /usr/share/locale:
     bind: $SNAP/usr/share/locale
 
+plugs:
+  read-snap-data:
+    interface: system-files
```
kenvandine, line 19: We could use the desktop-launch interface to get access to the icon.  Even though we aren't currently using that interface to launch snaps, it does provides access to the desktop files and icons.

sergio-costas: I'll test that.

sergio-costas: Mmm... it's not enough... the icon paths that I receive are inside the snaps, while "desktop-launch" only gives access to /var/lib/snapd/desktop/icons/...

kenvandine: The desktop-launch interface does include "/snap/*/*/** r," and I think it's more compelling to get permission to plug desktop-launch than system-files.

https://github.com/snapcore/snapd/blob/master/interfaces/builtin/desktop_launch.go#L46

sergio-costas: Odd... it works, but I think that the last time that I tested it, it didn't... Maybe I forgot to connect the plug... Anyway, change done.

kenvandine (approved):
Thanks for the change to use desktop-launch, I think this is good now.

PR #50, opened 2023-03-16
Updated the refresh message

sergio-costas:


sergio-costas:
@Mejans BTW: if you want to update the translation...

kenvandine (approved):
LGTM

PR #53, opened 2023-04-05
Use system-files interface

sergio-costas:
Replace the launch interface with the system-files one, as requested.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -14,6 +14,12 @@ layout:
   /usr/share/locale:
     bind: $SNAP/usr/share/locale
 
+plugs:
+  read-snap-data:
+    interface: system-files
+    read:
+      - /snap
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -14,6 +14,12 @@ layout:
   /usr/share/locale:
     bind: $SNAP/usr/share/locale
 
+plugs:
+  read-snap-data:
```
kenvandine, line 18: I think the convention required for naming the plug says this should map the path exposed, which in this case would simply be "snap", but that does feel weird to me.  Can you please s/read-snap-data/read-snap/ and we'll see what they say?

sergio-costas: Done.

kenvandine (approved):
Thanks, this is good now.

PR #54, opened 2023-04-05
Make the window bigger if needed, but not smaller

sergio-costas:
This MR makes the window bigger if the text requires more space, but once it has grown, it will keep the new size, no matter if new texts are smaller. This avoids the ugly effect of a window increasing and decreasing its size.

### DIFF
--- a/src/refresh_status.c
+++ b/src/refresh_status.c
@@ -35,6 +35,23 @@ void on_hide_clicked(GtkButton *button, RefreshState *state) {
   refresh_state_free(state);
 }
 
+void on_main_window_size_allocate(GtkApplicationWindow *self,
+                                  GtkAllocation *allocation,
+                                  RefreshState *state) {
+  gboolean modified = FALSE;
+  if (allocation->width > state->width) {
+    modified = TRUE;
+    state->width = allocation->width;
+  }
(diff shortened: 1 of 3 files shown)


sergio-costas:
@kenvandine Please, review this when you have some spare time. It's not critical, just an aesthetic detail.

src/refresh_status.c
```diff
@@ -35,6 +35,24 @@ void on_hide_clicked(GtkButton *button, RefreshState *state) {
...
+  if (allocation->width > state->width) {
+    modified = TRUE;
+    state->width = allocation->width;
+  }
+  if (allocation->height > state->height) {
+    modified = TRUE;
+    state->height = allocation->height;
+  }
+  if (modified) {
+    printf("Change size\n");
```
kenvandine, line 51: Let's drop the print statement.

kenvandine (changes requested):
Let's drop the printf.

src/refresh_status.c
sergio-costas: Oops... yeah, my fault... debugging notes O:-)

kenvandine (approved):
Looks good, thanks

PR #56, opened 2023-04-17
Allow to wait for a change in lock file

sergio-costas:
It adds a new option to wait for the lock file to be locked, and only then close the dialog when the file is unlocked.


sergio-costas:
@kenvandine Ken, can you review this when you have some spare time, please? It's a little change to allow to use a stripped-down version of the snapd code, to have a stopgap version as soon as possible, and then work on "nicer and more beauty" dialogs.

kenvandine (approved):
LGTM

PR #58, opened 2023-04-21
Add "desktop_file" option

sergio-costas:
This option allows to pass a desktop_file option, from where to extract the icon.


sergio-costas:
@kenvandine When you have some spare time can you review this, please?

kenvandine (approved):
Nice change!

PR #59, opened 2023-04-21
Ensure that the icon is visible only when set

sergio-costas:
By default the icon should be not visible, but when it is set, it must be changed to "visible". The previous patch made it invisible by default, but didn't make it visible when it is set. This patch fixes this.


sergio-costas:
@ken A little change that I forgot to include in the previous MR... (sorry).

kenvandine (approved):

PR #61, opened 2023-05-03
Migrate to core22 gtk4

sergio-costas:
Migrates snapd-desktop-integration to GTK4 (but still has to be ported to libadwaita).


sergio-costas:
@kenvandine Completed GTK4 port.

sergio-costas:
Porting to libadwaita should be quite straightforward: just replace one or two widgets, nothing else.

robert-ancell (approved):
LGTM

PR #77, opened 2024-06-04
Udeng-2978: put the right icon in the top left corner

sergio-costas:
Due to the fact that snap doesn't set a .desktop file for the snaps with the original name, but builds one with snap-name_app-name.desktop, the desktop identification usually fails. This makes GNotify to not be able to identify the .desktop file and set the right icon in the top-left corner of the notifications it sends.

imagen

This patch migrates the notification code to libnotify, which allows to bypass this problem. But the original code is kept  inside an #ifdef to allow to revert to it in the future, when this problem is fixed in snapd.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -137,7 +137,9 @@ static gboolean check_graphical_sessions(gpointer data) {
 
 static void do_startup(GObject *object, gpointer data) {
   GError *error = NULL;
-  notify_init("snapd-desktop-integration");
+#ifndef USE_GNOTIFY
+  notify_init("snapd-desktop-integration_snapd-desktop-integration");
+#endif
   client = snapd_client_new();
   refresh_monitor = sdi_refresh_monitor_new(G_APPLICATION(object));
   if (!sdi_refresh_monitor_start(refresh_monitor, &error)) {
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -178,8 +178,9 @@ static void show_pending_update_notification(SdiNotify *self,
                                   notification);
 }
 
-void show_simple_notification(SdiNotify *self, const gchar *title,
-                              const gchar *body, GIcon *icon, const gchar *id) {
+static void show_simple_notification(SdiNotify *self, const gchar *title,
+                                     const gchar *body, GIcon *icon,
+                                     const gchar *id) {
   g_autoptr(GNotification) notification = g_notification_new(title);
   g_notification_set_body(notification, body);
   if (icon != NULL) {
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -118,9 +118,10 @@ static void show_pending_update_notification(SdiNotify *self,
       notify_notification_new(title, body, icon_name);
   // don't use g_autoptr with the GVariant because it is consumed in set_hint
   GVariant *hint_icon = NULL;
-  if (icon_name != NULL)
+  if (icon_name != NULL) {
     hint_icon = g_variant_new_string(icon_name);
-  notify_notification_set_hint(notification, "image-path", hint_icon);
+    notify_notification_set_hint(notification, "image-path", hint_icon);
+  }
   notify_notification_add_action(notification, "app.close-notification",
                                  _("Close"), app_close_notification, NULL,
                                  NULL);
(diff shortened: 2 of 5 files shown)


sergio-costas:
@robert-ancell This is ready for reviewing too.

sergio-costas:
@robert-ancell BTW: Oliver commented to wait before building a new snap revision containing all these patches for refresh awareness, to test them in depth.

src/main.c
```diff
@@ -137,7 +137,9 @@ static gboolean check_graphical_sessions(gpointer data) {
 
 static void do_startup(GObject *object, gpointer data) {
   GError *error = NULL;
-  notify_init("snapd-desktop-integration");
+#ifndef USE_GNOTIFY
```
robert-ancell, line 140: Where does this define come from? Why do we not always use this?

src/sdi-notify.c
```diff
@@ -35,12 +36,126 @@ struct _SdiNotify {
...
+    g_signal_emit_by_name(self, "ignore-snap-event", apps[i]);
+}
+
 static GTimeSpan get_remaining_time(SnapdSnap *snap) {
   g_autoptr(GDateTime) proceed_time = snapd_snap_get_proceed_time(snap);
   g_autoptr(GDateTime) now = g_date_time_new_now_local();
   return g_date_time_difference(proceed_time, now);
 }
 
+static GVariant *get_snap_list(GSList *snaps) {
```
robert-ancell, line 55: You should document this returns a floating reference - It was unclear to me if this was leaking or not.

src/sdi-notify.c
```diff
@@ -35,12 +36,126 @@ struct _SdiNotify {
...
+static void show_simple_notification(SdiNotify *self, const gchar *title,
+                                     const gchar *body, GIcon *icon,
+                                     const gchar *id) {
+  g_autofree gchar *icon_name = get_icon_name_from_gicon(icon);
+  // Don't use g_autoptr because it must survive for the actions
+  NotifyNotification *notification =
+      notify_notification_new(title, body, icon_name);
+
+  if (icon_name != NULL) {
+    // don't use g_autoptr with the GVariant because it is consumed in set_hint
```
robert-ancell, line 148: Could drop this comment by using one line: `notify_notification_set_hint(notification, "image-path", g_variant_new_string(icon_name));`

src/sdi-notify.c
```diff
@@ -53,16 +168,11 @@ static void show_pending_update_notification(SdiNotify *self,
...
-  for (; snaps != NULL; snaps = snaps->next) {
-    SnapdSnap *snap = (SnapdSnap *)snaps->data;
-    g_variant_builder_add(builder, "s", snapd_snap_get_name(snap));
-  }
-  GVariant *values = g_variant_builder_end(builder);
-  g_notification_add_button_with_target_value(notification, "Ignore",
+  g_notification_add_button_with_target(notification, _("Close"),
+                                        "app.close-notification", "s",
+                                        "pending-update");
+  GVariant *values = get_snap_list(snaps);
```
robert-ancell, line 174: Could use a single line again, to avoid the confusion about the floating reference.

src/sdi-notify.c
sergio-costas: Fixed.

src/sdi-notify.c
sergio-costas: Again, good point. Done too.

src/sdi-notify.c
sergio-costas: You are right. But I preferred to sink the reference before returning it because that makes the code simpler to understand. What do you think?

src/main.c
sergio-costas: As this patch description says, there is a problem with gnotify and snapd, so it can't identify the right .desktop file and doesn't set the top-left icon in the notifications. Changing to libnotify instead allows to set that icon, which is what this patch does.

I added an screen capture with the problem in the top description.

src/sdi-notify.c
robert-ancell: I was going to suggest that - the floating references just add confusion IMHO and using normal references is much clearer.

src/main.c
robert-ancell: It's probably useful to add a comment here indicating why this is optional / not working.

sergio-costas: Added.

robert-ancell (approved):
Looks good. Note to self we are not merging this yet though.

soumyaDghosh:
It's the same in firefox mpris :-(

image

sergio-costas:
@soumyaDghosh Good to know, thanks!

sergio-costas:
@robert-ancell BTW, why shouldn't we merge this yet?

PR #79, opened 2024-06-17
Add desktop-launch in the plug list

sergio-costas:
The *desktop-launch* interface is required to be able to read the application icon and the visible name. The automatic connection is already requested at https://forum.snapcraft.io/t/request-for-desktop-launch-interface-for-snapd-desktop-integration/40634

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -35,6 +35,7 @@ apps:
       - login-session-observe
       - snap
       - snap-refresh-observe
+      - desktop-launch
 
 slots:
   snapd-desktop-integration:


sergio-costas:
@robert-ancell This one is ready for review.

olivercalder:
If `desktop-launch` is being added, is `system-files` still required? The `desktop-launch` interface grants access to `/snap/*/*/**`, and the only other reason I can think of for needing `/snap/**` would be to identify which snaps are installed. But I believe this is possible using one of the other interfaces (`snap-refresh-observe`)?

olivercalder:
Also, see: https://github.com/snapcore/snapd/pull/14168

This could potentially allow accessing installed snaps' icons via `/v2/icons/{name}/icon` without needing `desktop-launch`.

snap/snapcraft.yaml
```diff
@@ -35,6 +35,7 @@ apps:
       - login-session-observe
       - snap
       - snap-refresh-observe
+      - desktop-launch
```
olivercalder, line 38: Is there any need for functionality granted by `desktop-launch` other than reading `.desktop` files and icons? AFAICT the `snap` plug already grants all access required for this since it is `system-files` read access to `/snap`, which I believe grants access to all subdirectories as well.

snap/snapcraft.yaml
```diff
@@ -35,6 +35,7 @@ apps:
       - login-session-observe
       - snap
```
olivercalder, line 36: Is there anything other than reading icon files that `system-files` is needed for? Otherwise, it seems that `desktop-launch` grants everything required, since it grants `/snap/*/*/**`.

olivercalder (commented):
I don't believe both `system-files` and `desktop-launch` should be needed simultaneously. What is the exact functionality required of reading things in `/snap`?

Better yet, if you just need icons and visible names, I'd like to provide a cleaner way directly from snapd which doesn't require superprivileged interfaces and digging around the filesystem :)

snap/snapcraft.yaml
sergio-costas: No, it's a little bit more complex than that. The "desktop-legacy" and "unity7" plugs allow to read the folder with the .desktop files, but... they also forbid reading the contents of the .desktop files. They do this using AppArmor rules. The system-files read access is done with AppArmor rules too, and here is the problem: any deny is stronger than any allow, so it doesn't matter what you do, if you have "desktop-legacy" (which is automatically connected if you use the "desktop" plug, which snapd-desktop-integration requires to be able to show Gtk windows and notifications) or "unity7" (which we need too to be able to interact with the Dock), you won't be able to read the .desktop files contents, which we require to get the "pretty name" and the icon.

But "desktop-launch" does grant access to the contents, thanks to a patch that was merged in snapd some days ago (https://github.com/snapcore/snapd/pull/14061), because if it is connected, it will remove the lines that deny access to the .desktop files. At the beginning I did it there to avoid creating a new interface, but now that we also want to open the snap-center and the application itself after refresh, this interface is absolutely required.

snap/snapcraft.yaml
sergio-costas: Mmm... good point. So maybe we can remove this one... I'll check what does this one.

sergio-costas: Yes, it seems that it can be safely removed.

sergio-costas: Removed and rebased.

snap/snapcraft.yaml
olivercalder: Needing it to launch the snap store and application makes good sense, thanks! I don't see `desktop-legacy` or `unity7` in the plug list, so I was confused why those deny rules would be emitted.

olivercalder (approved):
LGTM, thank you!

sergio-costas:
@olivercalder `desktop-legacy` is connected "in the background" when using `desktop`. `unity7` isn't used here, but is required for the PR that shows the progress bar in the Dock: #80

sergio-costas:
@olivercalder I did some extra tests and it seems that the `desktop-launch` doesn't completely replaces the 'system-files' plug as defined here, because it only allows to read *inside* the snaps, but doesn't allow to see what snaps are available... so I'm now not so sure if we should remove it... @robert-ancell opinions?

olivercalder:
> @olivercalder I did some extra tests and it seems that the `desktop-launch` doesn't completely replaces the 'system-files' plug as defined here, because it only allows to read _inside_ the snaps, but doesn't allow to see what snaps are available... (this is: `ls /snap` only works with the `system-files` plug connected), so I'm now not so sure if we should remove it... @robert-ancell opinions?

Hmm, that is unfortunate. I feel like there ought to be a less permissive way to tell what snaps are installed, but if `desktop-launch` already grants `/snap/*/*/**`, granting `/snap` is really nothing more than snap names. So this is probably the cleanest solution. Perhaps @robert-ancell or someone else knows the idiomatic way to check installed snaps?

PR #81, opened 2024-06-19
Full raa

sergio-costas:
This is just the code with all the pending PRs, to do a snap for testing the new RAA.


sergio-costas:
Closing because it's not needed anymore

PR #82, opened 2024-07-02
Require last version of snapd-glib

sergio-costas:
Several changes done in snapd-desktop-integration now requires some extra API methods from snapd-glib. This PR updates the required version.


seb128 (approved):
Thanks, indeed it requires the newest version

PR #83, opened 2024-07-02
Update the list of files to translate

seb128:


sergio-costas (approved):
My fault, forgot to add those. Thanks!

PR #84, opened 2024-07-05
Launch the store from refresh notification

sergio-costas:
The UX team asked for allowing to launch the store directly from a pending refresh notification, to allow the user to see what snaps are pending. Also recommended a change in the text of the other button in the notifications.

Captura desde 2024-07-05 19-23-17

EDIT: now it also launches the updated snap when clicking in the "Snap is updated" notification.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -36,6 +36,7 @@ apps:
       - snap
       - snap-refresh-observe
       - unity7
+      - desktop-launch
 
 slots:
   snapd-desktop-integration:
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -18,8 +18,7 @@
 #define SECONDS_IN_A_DAY 86400
 #define SECONDS_IN_AN_HOUR 3600
 #define SECONDS_IN_A_MINUTE 60
-#define SNAP_STORE                                                             \
-  "/var/lib/snapd/desktop/applications/snap-store_snap-store.desktop"
+#define SNAP_STORE "snap-store_snap-store.desktop"
 
 #include <gio/gdesktopappinfo.h>
 #include <glib/gi18n.h>
@@ -319,7 +318,7 @@ GApplication *sdi_notify_get_application(SdiNotify *self) {
 static void sdi_notify_action_show_updates(GActionGroup *action_group,
                                            GVariant *str_data,
                                            SdiNotify *self) {
-  sdi_launch_desktop(self->application, "snap-store_snap-store.desktop");
+  sdi_launch_desktop(self->application, SNAP_STORE);
 }
 
 static void set_actions(SdiNotify *self) {
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -38,20 +38,22 @@ struct _SdiNotify {
 
 G_DEFINE_TYPE(SdiNotify, sdi_notify, G_TYPE_OBJECT)
 
-static void sdi_notify_action_ignore(GActionGroup *action_group,
-                                     GVariant *app_list, SdiNotify *self) {
-  gsize len;
-  g_autofree gchar **apps = (gchar **)g_variant_get_strv(app_list, &len);
-
-  g_return_if_fail(apps != NULL);
-  for (gsize i = 0; i < len; i++)
-    g_signal_emit_by_name(self, "ignore-snap-event", apps[i]);
-}
-
 static GTimeSpan get_remaining_time(SnapdSnap *snap) {
   g_autoptr(GDateTime) proceed_time = snapd_snap_get_proceed_time(snap);
   g_autoptr(GDateTime) now = g_date_time_new_now_local();
   return g_date_time_difference(proceed_time, now);
+}
+
+static void show_updates(SdiNotify *self) {
+  if (!sdi_launch_desktop(self->application, SNAP_STORE_UPDATES)) {
+    sdi_launch_desktop(self->application, SNAP_STORE);
+  }
+}
+
+static void sdi_notify_action_show_updates(GActionGroup *action_group,
+                                           GVariant *str_data,
+                                           SdiNotify *self) {
+  show_updates(self);
 }
 
 static GVariant *get_snap_list(GSList *snaps) {
@@ -64,6 +66,16 @@ static GVariant *get_snap_list(GSList *snaps) {
     g_variant_builder_add(builder, "s", snapd_snap_get_name(snap));
   }
   return g_variant_ref_sink(g_variant_builder_end(builder));
+}
+
+static void sdi_notify_action_ignore(GActionGroup *action_group,
+                                     GVariant *app_list, SdiNotify *self) {
+  gsize len;
+  g_autofree gchar **apps = (gchar **)g_variant_get_strv(app_list, &len);
+
+  g_return_if_fail(apps != NULL);
+  for (gsize i = 0; i < len; i++)
+    g_signal_emit_by_name(self, "ignore-snap-event", apps[i]);
 }
 
 // Currently, due to the way Snapd creates the .desktop files, the notifications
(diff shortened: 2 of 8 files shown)


robert-ancell (approved):
Looks good, can land after the rebase.

src/sdi-notify.c
```diff
@@ -18,6 +18,8 @@
 #define SECONDS_IN_A_DAY 86400
 #define SECONDS_IN_AN_HOUR 3600
 #define SECONDS_IN_A_MINUTE 60
+#define SNAP_STORE                                                             \
+  "/var/lib/snapd/desktop/applications/snap-store_snap-store.desktop"
```
robert-ancell, line 22: This doesn't seem to be used.

sergio-costas: Fixed, and adapted to the new format.

sergio-costas:
Had to do some extra changes after the rebase.

robert-ancell:
Ugh, needs rebase again :)

sergio-costas:
@robert-ancell Rebased.

snap/snapcraft.yaml
```diff
@@ -36,6 +36,7 @@ apps:
       - snap
```
olivercalder, line 36: I don't believe using `system-files` to read `/snap` is necessary anymore. `desktop-launch` grants access to `/snap/*/*/**`, so all `.desktop` and icon files are accessible (though for the record, the `.desktop` files in `/var/lib/snapd/desktop/applications` are cleaned up and more consistent to find and use, with absolute icon paths, etc.).

The `snap-refresh-observe` interface grants access to `/v2/snaps`, which allows getting information about all installed snaps. So I don't believe there's any reason to include this `snap` plug which uses `system-files`.

What, if any, information do you need to extract from `.desktop` files? My impulse is that, if possible, any information should be retrieved from `/v2/snaps/{snap}` first, rather than searching through a desktop file in `/snap/{snap}/current/meta/gui`. The exception to this is icon files, since most snaps do not package icon files in the expected location, so the `/v2/icons/{snap}/icon` endpoint rarely works, and the only way to find the local icon for an app is to look in the app's `.desktop` file.

If icons are the only reason, I would be in favor of adding a new snapd API endpoint, something like `/v2/icons/{snap}/icon/{app}`, which retrieves the local app icon. That way hopefully different projects don't need to dig through `.desktop` files anymore.

olivercalder (commented):
Question about interface usage

snap/snapcraft.yaml
sergio-costas: I need to read both the icon and the "Visible name", because it is better "Telegram Desktop" than "telegram-desktop".

sergio-costas: About removing the `system-files` interface... One problem at a time, please :-)

olivercalder: Would the `"title"` field from `/v2/snaps/{name}` work instead?

```
me@hostname:~$ curl -sS --unix-socket /run/snapd.socket http://localhost/v2/snaps/telegram-desktop | jq
{
  "type": "sync",
  "status-code": 200,
  "status": "OK",
  "result": {
    "id": "jpdoHqMd90M39DdH2JSZbeQxipqHtwLo",
    "title": "Telegram Desktop",
...
```

This is granted by `snap-refresh-observe`, without needing to search the `.desktop` file.

On the other hand, this is the snap name, rather than the app name. And there's no equivalent for each `app`:

```
    "apps": [
      {
        "snap": "telegram-desktop",
        "name": "telegram-desktop",
        "desktop-file": "/var/lib/snapd/desktop/applications/telegram-desktop_telegram-desktop.desktop",
        "common-id": "org.telegram.desktop"
      }
    ],
```

So perhaps reading the `.desktop` file is the best solution, at the moment (though I think the apps fields ought to have the "visible name" in addition to the `"name"` field, which is really just the internal `"app"` name, rather than the application's nice name.

olivercalder: The problem with using `.desktop` files is that they're per-app. For example, Libreoffice has many apps, and one does not usually run the one called `libreoffice`:

```
me@hostname:~$ grep '^Name=L' /var/lib/snapd/desktop/applications/libreoffice*.desktop
/var/lib/snapd/desktop/applications/libreoffice_base.desktop:Name=LibreOffice 24.2 Base
/var/lib/snapd/desktop/applications/libreoffice_calc.desktop:Name=LibreOffice 24.2 Calc
/var/lib/snapd/desktop/applications/libreoffice_draw.desktop:Name=LibreOffice 24.2 Draw
/var/lib/snapd/desktop/applications/libreoffice_impress.desktop:Name=LibreOffice 24.2 Impress
/var/lib/snapd/desktop/applications/libreoffice_libreoffice.desktop:Name=LibreOffice 24.2
/var/lib/snapd/desktop/applications/libreoffice_math.desktop:Name=LibreOffice 24.2 Math
/var/lib/snapd/desktop/applications/libreoffice_writer.desktop:Name=LibreOffice 24.2 Writer
me@hostname:~$ grep '^Name=L' /snap/libreoffice/current/**/*.desktop                  
/snap/libreoffice/current/lib/libreoffice/share/xdg/base.desktop:Name=LibreOffice 24.2 Base
/snap/libreoffice/current/lib/libreoffice/share/xdg/calc.desktop:Name=LibreOffice 24.2 Calc
/snap/libreoffice/current/lib/libreoffice/share/xdg/draw.desktop:Name=LibreOffice 24.2 Draw
/snap/libreoffice/current/lib/libreoffice/share/xdg/impress.desktop:Name=LibreOffice 24.2 Impress
/snap/libreoffice/current/lib/libreoffice/share/xdg/math.desktop:Name=LibreOffice 24.2 Math
/snap/libreoffice/current/lib/libreoffice/share/xdg/startcenter.desktop:Name=LibreOffice 24.2
/snap/libreoffice/current/lib/libreoffice/share/xdg/writer.desktop:Name=LibreOffice 24.2 Writer
/snap/libreoffice/current/lib/libreoffice/share/xdg/xsltfilter.desktop:Name=LibreOffice 24.2 XSLT based filters
/snap/libreoffice/current/meta/gui/base.desktop:Name=LibreOffice 24.2 Base
/snap/libreoffice/current/meta/gui/calc.desktop:Name=LibreOffice 24.2 Calc
/snap/libreoffice/current/meta/gui/draw.desktop:Name=LibreOffice 24.2 Draw
/snap/libreoffice/current/meta/gui/impress.desktop:Name=LibreOffice 24.2 Impress
/snap/libreoffice/current/meta/gui/libreoffice.desktop:Name=LibreOffice 24.2
/snap/libreoffice/current/meta/gui/math.desktop:Name=LibreOffice 24.2 Math
/snap/libreoffice/current/meta/gui/writer.desktop:Name=LibreOffice 24.2 Writer
```

If you want the visible name of the snap itself, rather than a particular app, I would strongly recommend against using a value from a `.desktop` file, as those are app-specific.

For example, one could create a snap named `foo` which has apps `bar` and `baz`, with `bar.desktop` and `baz.desktop` in `/snaps/foo/current/meta/gui` containing `Name=Bar` and `Name=Baz`, respectively. Or what if the `libreoffice` snap did not include a `libreoffice` app. There's no way to decide which of those names should be used when referring to the snap itself, other than the convention that there is often a single application per snap, or that if there are multiple, there is one app matching the snap name. But this is not guaranteed.

sergio-costas: It looks good...

olivercalder: I think I initially misunderstood though. Do you want the name/icon for the snap or for the app?

sergio-costas: The point is that, sometimes, the icon in the snap doesn't match the icon for the app, so that's why I preferred to use the one from the .desktop. Anyway, you are right in that when there are several applications (like Libreoffice) this can be problematic.

Also... Do all snaps have a "pretty name"? (I mean: the "title" field).

sergio-costas: Anyway, I agree in that searching the .desktop files is an ugly solution.

sergio-costas: I've been checking and there are a lot of snaps that lack the "title" entry, so what do you think about trying the "title" entry, and if it isn't available, use the heuristics for the .desktop file? And the same for the icon.

olivercalder: I completely agree, that's a great approach. It's a bit more work/complexity for you though, unfortunately :) but I think it's probably the best we can do.

I just checked, and every snap listed in the store has a `"title"` field, but they're often just the snap name:
```
me@hostname:~$ for s in $(cat /var/cache/snapd/names); do echo "$s:\t$(curl -f -sS --unix-socket /run/snapd.socket http://localhost/v2/find\?name=$s | jq '.result[] | .title')" | tee -a /tmp/snap_titles.txt ; done
```
Results here: https://bucket.calder.dev/snap_titles.txt

Even `firefox` just has `"firefox"` as its title, unfortunately. The only snaps without `"title"` fields are unlisted: `bare` and `gtk-common-themes`, at least on my system:
```
me@hostname:~$ for s in $(snap list | tail +2 | cut -d ' ' -f 1); do echo "$s:\t$(curl -f -sS --unix-socket /run/snapd.socket http://localhost/v2/snaps/$s | jq .result.title)" ; done | grep null   
bare:	null
gtk-common-themes:	null
```

So my best guess about priority/heuristics would be:

1. If `"title"` exists and starts with capital letter, use it
2. If one `.desktop` exists or multiple exist and one exactly matches the snap name, use the `Name=` from it (and icon)
3. If `"title"` exists, use it
4. If multiple `.desktop` files exist, try to choose the best one and use the `Name=` form it
5. Use the snap name as a last resort (should only be necessary for unlisted snaps)

sergio-costas: Looks sensible. I'll do it when I complete the other tasks.

sergio-costas:
@robert-ancell Should be ready for merging.

sergio-costas:
(I don't have write permissions)

PR #85, opened 2024-07-11
Cleanup code

sergio-costas:
Use _new and _free functions for IgnoreNotifyData structs.

### DIFF
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -97,6 +97,22 @@ typedef struct {
   GVariant *snaps;
 } IgnoreNotifyData;
 
+static IgnoreNotifyData *ignore_notify_data_new(SdiNotify *self,
+                                                GVariant *snaps) {
+  IgnoreNotifyData *data = g_malloc0(sizeof(IgnoreNotifyData));
+  data->self = g_object_ref(self);
+  data->snaps = g_variant_ref(snaps);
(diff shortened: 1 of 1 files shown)


src/sdi-notify.c
```diff
@@ -132,9 +146,9 @@ static void show_pending_update_notification(SdiNotify *self,
                                  NULL);
   notify_notification_add_action(notification, "default", _("Close"),
                                  app_close_notification, NULL, NULL);
-  IgnoreNotifyData *data = g_malloc(sizeof(IgnoreNotifyData));
-  data->self = self;
-  data->snaps = get_snap_list(snaps);
+  g_autoptr(GVariant) snap_list = get_snap_list(snaps);
+  // The memory block of data will be freed in the callback
```
robert-ancell, line 150: This is sometimes clearer by just doing it inline, i.e.
```
notify_notification_add_action(notification, "app.ignore-notification",
                                 _("Ignore"), app_ignore_snaps_notification,
                                 ignore_notify_data_new(self, snap_list), NULL);
```

Also, you can set the last argument to `ignore_notify_data_free` and then it doesn't have to be unreffed in the callback.

sergio-costas: good point. Done.

robert-ancell (approved):

PR #87, opened 2024-07-12
Fix crash when showing several progress bars

sergio-costas:
When calling foreach in the dictionary, g_object_ref() is called once on self, but in the callback it is assigned to a g_autoptr in each call. If there is only one entry in the dictionary, it works, but if there are several, the second time the callback is called, self will be freed again, and the program crashes.


sergio-costas:
@robert-ancell Can you review and merge this when you have some spare time, please?

robert-ancell (approved):
LGTM

PR #88, opened 2024-07-19
Fix odd progress bar

sergio-costas:
Sometimes, an incorrect value with the number of tasks is sent, and the progress bar shows an odd appearance. This patch fixes it.


sergio-costas:
@robert-ancell Can you review this one when you have some spare time, please?

robert-ancell (approved):
LGTM

PR #89, opened 2024-07-22
Fix cancelled refreshes

sergio-costas:
When a refresh was cancelled (for example, because the snap was launched before it has completed the refresh), the snap is reverted, but the current code fails to detect correctly this, and snapd-desktop-integration could be left in an invalid state, asking for already completed changes over and over again.

This patch fixes this bug.

### DIFF
--- a/src/sdi-refresh-monitor.c
+++ b/src/sdi-refresh-monitor.c
@@ -267,12 +267,12 @@ static void refresh_change(gpointer p) {
 static gboolean status_is_done(const gchar *status) {
   gboolean done = g_str_equal(status, "Done") || g_str_equal(status, "Abort") ||
                   g_str_equal(status, "Error") || g_str_equal(status, "Hold") ||
-                  g_str_equal(status, "Wait");
+                  g_str_equal(status, "Wait") || g_str_equal(status, "Undone");
   return done;
 }
 
 static void update_inhibited_snaps(SdiRefreshMonitor *self, SnapdChange *change,
-                                   gboolean done, gboolean hold) {
+                                   gboolean done, gboolean cancelled) {
   SnapdAutorefreshChangeData *change_data =
       SNAPD_AUTOREFRESH_CHANGE_DATA(snapd_change_get_data(change));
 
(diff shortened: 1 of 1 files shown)


sergio-costas:
@robert-ancell When you have some spare time, can you review this too? The true patch is the last commit only.

robert-ancell:
Does this need a merge/rebase? The diff seems too large.

sergio-costas:
@robert-ancell Fixed.

src/sdi-refresh-monitor.c
```diff
@@ -439,8 +444,15 @@ static void manage_change_update(SnapdClient *source, GAsyncResult *res,
...
 
-  gboolean done = g_str_equal(snapd_change_get_status(change), "Done");
-  gboolean cancelled = cancelled_change_status(snapd_change_get_status(change));
+  const gchar *change_status = snapd_change_get_status(change);
+
+  gboolean done = g_str_equal(change_status, "Done");
+  gboolean cancelled = cancelled_change_status(change_status);
+  gboolean valid_do = valid_working_change_status(change_status);
+  if (!(valid_do || cancelled)) {
+    g_print("Unknown state %s\n", change_status);
```
robert-ancell, line 453: g_warning or g_debug instead?

robert-ancell: Also the text could probably be more specific.

sergio-costas: Good point.

sergio-costas:
Rebased again.

src/sdi-refresh-monitor.c
```diff
@@ -440,8 +445,15 @@ static void manage_change_update(SnapdClient *source, GAsyncResult *res,
...
 
-  gboolean done = g_str_equal(snapd_change_get_status(change), "Done");
-  gboolean cancelled = cancelled_change_status(snapd_change_get_status(change));
+  const gchar *change_status = snapd_change_get_status(change);
+
+  gboolean done = g_str_equal(change_status, "Done");
+  gboolean cancelled = cancelled_change_status(change_status);
+  gboolean valid_do = valid_working_change_status(change_status);
+  if (!(valid_do || cancelled)) {
+    g_debug("Unknown change status %s\n", change_status);
```
robert-ancell, line 454: Don't need the newline at the end - g_debug adds this already.

sergio-costas: Fixed.

robert-ancell (approved):

PR #90, opened 2024-07-23
Don't launch several "threads" for same change

sergio-costas:
When a change is processed, a timeout is set to continue monitoring it. Unfortunately, when a change is sent again to the program, a new timeout is set, thus having two "pseudothreads" monitoring the same change.

This patch checks if there is already a timeout for a change, and avoids creating a new timeout in that case, because the current one will already manage new changes.

### DIFF
--- a/src/sdi-refresh-monitor.c
+++ b/src/sdi-refresh-monitor.c
@@ -257,6 +257,7 @@ static void show_snap_completed(GObject *source, GAsyncResult *res,
 
 static void refresh_change(gpointer p) {
   g_autoptr(SnapRefreshData) data = p;
+  g_hash_table_remove(data->self->changes, data->change_id);
   snapd_client_get_change_async(data->self->client, data->change_id, NULL,
                                 (GAsyncReadyCallback)manage_change_update,
                                 g_object_ref(data->self));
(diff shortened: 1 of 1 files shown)


sergio-costas:
@robert-ancell Can you review this when you have some spare time, please?

src/sdi-refresh-monitor.c
```diff
@@ -545,6 +548,7 @@ static void sdi_refresh_monitor_dispose(GObject *object) {
   }
   g_clear_pointer(&self->snaps, g_hash_table_unref);
   g_clear_object(&self->client);
+  g_clear_pointer(&self->changes, g_hash_table_unref);
```
robert-ancell, line 551: You probably should be calling g_source_remove on each value, which might be doable by setting a function in g_hash_table_new_full below. This is an existing bug, but is solvable now you are storing the ID.

robert-ancell (approved):
LGTM

src/sdi-refresh-monitor.c
sergio-costas: In fact, since the timeout callback is called with a count_ref pointer to "self", this can be called only when there are no pending timeouts. Also, since I have to remove the change in the timeout callback, in that moment the source is already removed, so using new_full would produce an error (or, at least, a warning).

sergio-costas: I can add a comment to explain this, if you want.

PR #93, opened 2024-09-10
Show updates when clicking on notification

sergio-costas:
When a notification with pending updates is shown, clicking on it should open the store with the available updates.

### DIFF
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -191,7 +191,11 @@ static void show_pending_update_notification(SdiNotify *self,
   notify_notification_add_action(notification, "app.show-updates",
                                  _("Show updates"), app_show_updates,
                                  g_object_ref(self), g_object_unref);
-  notify_notification_add_action(notification, "default", _("Close"),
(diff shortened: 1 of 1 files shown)


src/sdi-notify.c
```diff
@@ -192,7 +192,8 @@ static void show_pending_update_notification(SdiNotify *self,
                                  _("Show updates"), app_show_updates,
                                  g_object_ref(self), g_object_unref);
   notify_notification_add_action(notification, "default", _("Close"),
-                                 app_close_notification, NULL, NULL);
+                                 app_show_updates, g_object_ref(self),
+                                 g_object_unref);
```
robert-ancell, line 196: It seems quite confusing that both "Show updates" and "Close" do the same thing - It might be worth adding a comment here.

robert-ancell (approved):
LGTM

src/sdi-notify.c
sergio-costas: Done.

PR #94, opened 2024-09-16
Fix theme tests

sergio-costas:
The tests folder contains a test for theme installation, but it fails.

This patch fixes it.


robert-ancell (approved):
LGTM

PR #95, opened 2024-09-18
First code refactor for adding unitary tests

sergio-costas:
Some code refactoring with the aim of preparing it for adding unitary tests, and also to clean the code and make it more maintainable.

The first commit moves all the code that checks if we are in a graphical session into their own file, removing it from `main`.

The second commit replaces global variables into local ones wherever possible.

The third commit moves all the snapd-notices-monitor code into its own object. This object can be seen as a "super-notices-monitor", which hides when the notices socket is closed (for example, because snapd has been updated), thus keeping the illusion that the connection lasts forever.


src/sdi-refresh-monitor.c
```diff
@@ -618,9 +575,8 @@ void sdi_refresh_monitor_init(SdiRefreshMonitor *self) {
   self->refreshing_snap_list = g_hash_table_new_full(
       g_str_hash, g_str_equal, g_free, free_progress_task_data);
   self->client = snapd_client_new();
-  if (check_is_running_in_snap())
```
robert-ancell, line 621: Why did you remove this function? It makes the check more readable.

src/sdi-snapd-monitor.c
```diff
@@ -0,0 +1,105 @@
...
+               NULL, NULL, NULL, G_TYPE_NONE, 2, SNAPD_TYPE_NOTICE,
+               G_TYPE_BOOLEAN);
+}
+
+SdiSnapdMonitor *sdi_snapd_monitor_new() {
+  return g_object_new(SDI_TYPE_SNAPD_MONITOR, NULL);
+}
+
+gboolean sdi_snapd_monitor_start(SdiSnapdMonitor *self) {
+
```
robert-ancell, line 100: nit: empty line

src/main.c
```diff
@@ -58,15 +63,10 @@ static void do_startup(GObject *object, gpointer data) {
 
 static void do_activate(GObject *object, gpointer data) {
```
robert-ancell, line 67: Can update this to `do_activate(GtkApplication *app)` which will make the following code a little more readable.

robert-ancell:
This is somewhat hard to review, it would be easier as three PRs.

src/main.c
```diff
@@ -31,25 +31,30 @@
 #include "sdi-session-helpers.h"
 #include "sdi-theme-monitor.h"
 
-static SnapdClient *client = NULL;
-static GtkApplication *app = NULL;
-static SdiThemeMonitor *theme_monitor = NULL;
-static SdiRefreshMonitor *refresh_monitor = NULL;
```
robert-ancell, line 39: By making these local they are a bit harder to manage, as they are never unreffed (same issue currently but all in one place). This would be better managed as a GApplication which I worked on at some point but never finished. Not sure if the in-between step of making them local is worth it?

sergio-costas:
You are right: as three PRs will be easier. Changing it.

PR #96, opened 2024-09-19
Move session code into helper file

sergio-costas:
To reduce the code in main, the code used for checking if we are or not in a graphical session is moved into their own file. That's the only change done: move all the code for checking if the current session is a graphical one from `main` into its own source file.


src/meson.build
```diff
@@ -11,6 +11,7 @@ snapd_desktop_integration = executable(
   'sdi-refresh-dialog.c',
   'sdi-refresh-monitor.c',
   'sdi-theme-monitor.c',
+  'sdi-user-session-manager.c',
```
robert-ancell, line 14: Doesn't sound like a manager, as there is no class just one helper method. Not sure of a better name but perhaps something like `sdi-graphical-session-helper.c`.

robert-ancell (approved):
The name was unexpected but the code changes look good.

src/meson.build
sergio-costas: You are right. Changing it.

sergio-costas: Done.

PR #97, opened 2024-09-19
Move the snapd-notices-monitor logic into its own class and add unitary test

sergio-costas:
This PR moves the snapd notices monitor code into its own class/object. This new object will keep a connection permanently open, even if snapd closes it (for example, because it is updated), by opening a new one. It also re-sends each notice event with exactly the same format, thus allowing to connect to it with the same logic than the original snapd notices monitor.

There is no true new code, only moving the methods from sdi-refresh-monitor.c into its own class, and connecting both using signals. This makes the code less coupled. It also allows to add a fully automated unitary test for this functionality, that would be tested using CI every time the code is uploaded to Github.


sergio-costas:
This is basically moving code outside sdi-refresh-monitor.c into its own object in sdi-snapd-monitor.c, there are no changes in the logic.

src/main.c
```diff
@@ -30,19 +30,20 @@
...
 static SdiRefreshMonitor *refresh_monitor = NULL;
 static gboolean running = FALSE;
+static SdiSnapdMonitor *snapd_monitor = NULL;
 
-static gchar *snapd_socket_path = NULL;
-
-static GOptionEntry entries[] = {{"snapd-socket-path", 0, 0,
-                                  G_OPTION_ARG_FILENAME, &snapd_socket_path,
+static GOptionEntry entries[] = {{"snapd-socket-path", 0, G_OPTION_FLAG_NONE,
+                                  G_OPTION_ARG_FILENAME, &sdi_snapd_socket_path,
```
sergio-costas, line 46: @robert-ancell I need to have access to this parameter from the `helpers.c` file, to be able to create as many snapd-clients as needed using the right path. Which way do you think is better to do it?

* this way, exporting the `gchar *` variable from the `helpers.c` file and using it directly here
* keeping that string private to the `helpers.c` file and use a function to copy that parameter into the private variable, calling it in `do_startup()`
* moving the whole GOptionEntry variable into `helpers.c` and add a function, also in `helpers.c` that receives the GApplication variable and calls `g_application_add_main_option_entries(G_APPLICATION(app), entries);`

robert-ancell: `helpers.c` shouldn't have any state, so moving the variable into it suggests this is not good architecture. In general, having helpers indicates that these functions aren't in the right place - they should be owned by a class or module with clear purpose.

It seems to me there should be a `sdi_application_create_snapd_client` and a `SdiApplication` class that owns `snapd_socket_path`.

sergio-costas: Uf... creating a gobject singleton just for that is a little bit... I think that I'll create a "factory module".

sergio-costas: Thanks!

src/sdi-refresh-monitor.c
```diff
@@ -535,12 +532,10 @@ static void manage_refresh_inhibit(SnapdClient *source, GAsyncResult *res,
...
 
-static void notice_cb(GObject *object, SnapdNotice *notice, gboolean first_run,
-                      gpointer monitor) {
-  SdiRefreshMonitor *self = SDI_REFRESH_MONITOR(monitor);
-
-  g_autoptr(GHashTable) data = snapd_notice_get_last_data(notice);
-  g_autofree gchar *kind = g_strdup(g_hash_table_lookup(data, "kind"));
+void sdi_refresh_monitor_notice(SdiRefreshMonitor *self, SnapdNotice *notice,
+                                gboolean first_run, gpointer data) {
+  g_autoptr(GHashTable) notice_data = snapd_notice_get_last_data(notice);
```
robert-ancell, line 537: Should be `snapd_notice_get_last_data2` now?

src/sdi-snapd-client-factory.h
```diff
@@ -0,0 +1,23 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#pragma once
+
+#include <snapd-glib/snapd-glib.h>
+
+void sdi_snapd_client_factory_set_custom_path(gchar *path);
```
robert-ancell, line 22: `const gchar`?

src/sdi-snapd-client-factory.h
```diff
@@ -0,0 +1,23 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#pragma once
```
robert-ancell, line 18: Should this file use `G_BEGIN_DECLS`?

src/sdi-snapd-client-factory.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <snapd-glib/snapd-glib.h>
+
+void sdi_snapd_client_factory_set_custom_path(gchar *path);
+SnapdClient *sdi_snapd_client_factory_new_snapd_client(void);
```
robert-ancell, line 23: nit: Normally have an empty line between functions.

src/sdi-snapd-monitor.c
```diff
@@ -0,0 +1,114 @@
...
+                                           (GCallback)error_cb, self);
+}
+
+static void launch_snapd_monitor_after_error(SdiSnapdMonitor *self) {
+  configure_snapd_monitor(self);
+  snapd_notices_monitor_start(self->snapd_monitor, NULL);
+}
+
+static void error_cb(GObject *object, GError *error, SdiSnapdMonitor *self) {
+  g_print("Error %d; %s\n", error->code, error->message);
```
robert-ancell, line 71: Left over debugging? If kept should be g_debug and have a more descriptive message (i.e. which error?)

robert-ancell (approved):
Some small stuff but otherwise looks good.

src/sdi-snapd-client-factory.h
sergio-costas: Good point. I'll try to add it to clang-format.

src/sdi-snapd-client-factory.h
sergio-costas: Not really, because it really isn't a GObject. Should I add it anyway...?

src/sdi-snapd-client-factory.h
sergio-costas: Yes, good point.

src/sdi-refresh-monitor.c
sergio-costas: Yes, but since that was changed in a different patch, I kept it this way. Now that the PR has been merged, it has already been changed.

src/sdi-snapd-monitor.c
sergio-costas: Yes, absolutely right. Fixed, and also fixed others. Replaced all the g_prints with g_debugs and added extra debug info.

src/sdi-snapd-client-factory.h
sergio-costas: Done, and also in `helpers.h`.

PR #98, opened 2024-09-19
Clean exit

sergio-costas:
This patch manages the INT and TERM signals to cleanly quit the main loop and free all the used resources.


robert-ancell (approved):
LGTM

PR #99, opened 2024-09-25
Refactor window with progress dialogs

sergio-costas:
This PR moves the GTK window and dock progress bars management into their own classes, letting the `SdiRefreshMonitor` object to process the events from snapd and just send signals with the changes, signals that are connected to the new `SdiProgressWindow` and `SdiProgressDock` objects, which are responsible of creating a window if needed, update the progress bars in each element, and close a window if requested, for the former, and update the progress bar in the dock for the later.

This separation simplifies the code and, along with #97, allows to isolate the event management code, which will allow to create unitary tests for the process.


src/sdi-progress-dock.c
```diff
@@ -0,0 +1,97 @@
...
+  GObjectClass *gobject_class = G_OBJECT_CLASS(klass);
+
+  gobject_class->dispose = sdi_progress_dock_dispose;
+}
+
+static void sdi_progress_dock_init(SdiProgressDock *self) {}
+
+SdiProgressDock *sdi_progress_dock_new(GApplication *application) {
+  SdiProgressDock *self = g_object_new(SDI_TYPE_PROGRESS_DOCK, NULL);
+  self->application = g_object_ref(application);
```
aleasto, line 89: Isn't this one too many refs?

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+
+static void test_progress_bar(TestData *test) {
+  describe_test(test);
+  show_progress_window("A-SNAP", *app_list);
+
+  g_assert_cmpint(count_hash_childs(), ==, 1);
+  g_assert_cmpint(count_progress_childs(), ==, 1);
+  gint timeout_id = g_timeout_add(500, (GSourceFunc)set_progress_bar, "A-SNAP");
+  wait_for_click();
+  g_source_remove(timeout_id);
```
3v1n0, line 136: These are good cleanups, but in this case it's not needed to remove the source, I'd just use `g_timeout_add_once` and live with it.

Since you don't care about tracking what's happening once the callback is done.

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+    if (!g_app_info_should_show(app_info)) {
+      continue;
+    }
+    const gchar *command_line = g_app_info_get_commandline(app_info);
+    if (strstr(command_line, "/snapd/") != NULL) {
+      g_strv_builder_add(builder, g_app_info_get_id(p->data));
+    }
+  }
+  app_list = g_strv_builder_end(builder);
+  g_list_free_full(applist, g_object_unref);
```
3v1n0, line 295: Drop this

src/sdi-progress-window.c
```diff
@@ -0,0 +1,203 @@
...
+                                     gpointer data) {
+  SdiRefreshDialog *dialog =
+      (SdiRefreshDialog *)g_hash_table_lookup(self->dialogs, snap_name);
+  if (dialog != NULL) {
+    remove_dialog_from_main_window(self, dialog);
+  }
+}
+
+/**
+ * This callback should be connected to the `update-progress` signal from a
```
aleasto, line 155: Do you mean `refresh-progress`?
```suggestion
 * This callback should be connected to the `refresh-progress` signal from a
```

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+}
+
+static void do_activate(GApplication *app, gpointer data) {
+  create_window(app);
```
3v1n0, line 299: Make it to be `window = create_window(app)` to have things clearly managed.

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+    GAppInfo *app_info = G_APP_INFO(p->data);
+    if (!g_app_info_should_show(app_info)) {
+      continue;
+    }
+    const gchar *command_line = g_app_info_get_commandline(app_info);
+    if (strstr(command_line, "/snapd/") != NULL) {
+      g_strv_builder_add(builder, g_app_info_get_id(p->data));
+    }
+  }
+  app_list = g_strv_builder_end(builder);
```
3v1n0, line 294:
```suggestion
  g_assert_null (app_list);
  app_list = g_strv_builder_end(builder);
```

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+  for (GList *p = applist; p != NULL; p = p->next) {
+    GAppInfo *app_info = G_APP_INFO(p->data);
+    if (!g_app_info_should_show(app_info)) {
+      continue;
+    }
+    const gchar *command_line = g_app_info_get_commandline(app_info);
+    if (strstr(command_line, "/snapd/") != NULL) {
+      g_strv_builder_add(builder, g_app_info_get_id(p->data));
+    }
+  }
```
3v1n0, line 293: mnh, shouldn't we override in tests only the content of this key so that it surely contains a such desktop file?

It also should be a file that is provided by the test suite no?

src/main.c
```diff
@@ -95,6 +115,8 @@ static void do_shutdown(GObject *object, gpointer data) {
   g_clear_object(&client);
   g_clear_object(&theme_monitor);
   g_clear_object(&refresh_monitor);
+  g_clear_object(&progress_window);
```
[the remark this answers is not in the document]
sergio-costas: I agree... better for another PR.

src/sdi-progress-dock.c
```diff
@@ -0,0 +1,97 @@
...
+ * #sdi_refresh_monitor object. It will receive the total number of tasks and
+ * how many have been done, and if the task has been completed, and with that
+ * will update the progress bars in the dock.
+ */
+void sdi_progress_dock_update_progress(SdiProgressDock *self, gchar *snap_name,
+                                       GStrv desktop_files,
+                                       gchar *task_description,
+                                       guint done_tasks, guint total_tasks,
+                                       gboolean task_done, gpointer data) {
+  if ((desktop_files != NULL) && (total_tasks != 0)) {
```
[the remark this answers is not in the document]
sergio-costas: I personally really dislike that. I know that precedence can take care of those parentheses, but I'm of the opinion that being explicit is safer for reviewers, so those extra parameters just make the code more legible.

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+ */
+
+static void do_startup(GObject *object, gpointer data) {
+  progress_window = sdi_progress_window_new(G_APPLICATION(object));
+
+  // this gsettings key stores the path to the .desktop files of the programs
+  // in the dock. Since we need to test a progress bar in the dock, we need
+  // that list.
+  g_autoptr(GSettings) settings = g_settings_new("org.gnome.shell");
+  GList *applist = g_app_info_get_all();
```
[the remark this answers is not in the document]
sergio-costas: Good point. Anyway, I already run everything under valgrind. In fact, this isn't a leak because there was a g_list_free_full() after. But g_autolist() is certainly much more elegant.

tests/test-sdi-progress-window.c
sergio-costas: Done.

tests/test-sdi-progress-window.c
sergio-costas: Done.

tests/test-sdi-progress-window.c
sergio-costas: The problem is that we need to use .desktop files from snaps to ensure that everything works as expected. That's why the test is taking all the installed programs.

Anyway, remember that these tests are meant to be done "manually".

src/sdi-progress-dock.c
```diff
@@ -0,0 +1,97 @@
...
+    for (gchar **desktop_file = desktop_files; *desktop_file != NULL;
+         desktop_file++) {
+      // Update dock progress bar
+      g_autoptr(GVariantBuilder) builder =
+          g_variant_builder_new(G_VARIANT_TYPE("a{sv}"));
+      g_variant_builder_add(
+          builder, "{sv}", "progress",
+          g_variant_new_double(((gfloat)done_tasks) / ((gfloat)total_tasks)));
```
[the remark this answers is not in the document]
sergio-costas: That's done by clang-format. If I try to change it, the formatter restores it again.

src/sdi-progress-window.c
sergio-costas: Again, you are right. That was the old name of the signal. Clearly I forgot to change the references in the comments when I renamed it.

src/sdi-progress-dock.c
aleasto: Who owns the other reference? It's leaked at the end of this function so the memory will never be freed

src/main.c
```diff
@@ -76,6 +80,22 @@ static void do_startup(GObject *object, gpointer data) {
...
+                          (GCallback)sdi_progress_window_update_progress,
+                          progress_window, G_CONNECT_SWAPPED);
+  g_signal_connect_object(refresh_monitor, "end_refresh",
+                          (GCallback)sdi_progress_window_end_refresh,
+                          progress_window, G_CONNECT_SWAPPED);
+
+  progress_dock = sdi_progress_dock_new(G_APPLICATION(object));
+  g_signal_connect_object(refresh_monitor, "refresh_progress",
+                          (GCallback)sdi_progress_dock_update_progress,
+                          progress_dock, G_CONNECT_SWAPPED);
```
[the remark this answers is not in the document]
sergio-costas: Mmm... the odd thing is... how is it possible that it worked, if in one place I used - and in the others I used _ instead?
Fixed anyway.

src/sdi-progress-window.c
```diff
@@ -0,0 +1,203 @@
...
+  return self->dialogs;
+}
+
+GtkWindow *sdi_progress_window_get_window(SdiProgressWindow *self) {
+  return self->main_window;
+}
+
+#endif
+
+static gboolean check_parent_contains_child(GtkWidget *parent,
```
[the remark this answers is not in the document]
sergio-costas: Good point. Done.

src/sdi-progress-dock.c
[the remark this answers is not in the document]
sergio-costas: As before, I prefer to be a little bit more explicit to make simpler for any reviewer to identify what's being done... Anyway... what about something like this?

    g_variant_new_double(done_tasks / ((gfloat) total_tasks));

src/sdi-progress-dock.c
sergio-costas: The other reference is created in main function, at line 173, which has a g_autoptr() definition, and is passed as the first parameter when `do_startup()` is called.

sergio-costas: And is passed as a parameter to `sdi_progress_dock_new()` at line 94 in `main.c`.

src/sdi-progress-dock.c
```diff
@@ -0,0 +1,97 @@
...
+ * This callback should be connected to the `update-progress` signal from a
+ * #sdi_refresh_monitor object. It will receive the total number of tasks and
+ * how many have been done, and if the task has been completed, and with that
+ * will update the progress bars in the dock.
+ */
+void sdi_progress_dock_update_progress(SdiProgressDock *self, gchar *snap_name,
+                                       GStrv desktop_files,
+                                       gchar *task_description,
+                                       guint done_tasks, guint total_tasks,
+                                       gboolean task_done, gpointer data) {
```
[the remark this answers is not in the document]
sergio-costas: Yes, I know that it works, but... I really dislike doing that. Yes, *in theory*, in C, the caller is responsible for cleaning the stack, and since the parameters are stacked "in reverse", any extra parameter just won't be accessed, so they shouldn't be a problem... unless the calling convention says a different thing, like in the case of https://learn.microsoft.com/en-us/cpp/cpp/stdcall?view=msvc-170 (yes, I know that it is a little bit elaborated).

Anyway... the other cases do the same, so... for consistency...?

src/sdi-progress-dock.c
aleasto: Sorry, i misread the code

sergio-costas: No problem. That part is a little bit convoluted.

src/sdi-progress-dock.c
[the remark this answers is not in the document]
3v1n0: Ok, fair enough

src/main.c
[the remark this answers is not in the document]
3v1n0: Both works IIRC, but indeed I wanted to point out the same in my previous review.

Thanks for changing.

src/sdi-progress-dock.c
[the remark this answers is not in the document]
3v1n0: mh, personally it's harder to me :)

I would avoid adding parenthesis as they make things more complex to follow (as you're expecting having a sub-expression), but that's something to be defined as project code style I guess.

src/sdi-progress-dock.c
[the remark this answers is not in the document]
sergio-costas: Done.

tests/test-sdi-progress-window.c
sergio-costas: Or am I misunderstanding what you mean?

src/sdi-progress-dock.c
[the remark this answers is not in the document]
3v1n0: So, my point is that those functions are publicly accessible methods of the dock class, so they should be something that is potentially callable also directly, and not considered as signal callbacks.

If you want signal callbacks, then you can use local functions that adapt the callback arguments to the function that is being invoked.

Now, since in this case the signature match, we can just rely on `G_CALLBACK` doing its thing.

If you instead prefer to have a callback that has the whole signature, then use local functions to call them, but to me it looks just an unneeded repetition of code.

tests/test-sdi-progress-window.c
3v1n0: Well, I would like to see if the tests could be automated in a place where no snap is installed too, and so having local test  files that are equal to all the potential snap files makes it more reproducible and allows to check all the cases we care about and adjust the ones we're managing wrongly in case there's one.

sergio-costas: The window and dock classes should be quite "stable", so I'm not worried for these tests to be only "manually runable". The complex part is in `sdi-refresh-monitor`, and the idea with this and the previous refactors is, precisely, to fully isolate that code from Gtk and the desktop and be able to do automatic tests in that part.

src/sdi-progress-dock.c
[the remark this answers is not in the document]
sergio-costas: Well, after checking the GLib code, I see that the trick of relying on G_CALLBACK to remove the `data` parameter is used there, so... Ok.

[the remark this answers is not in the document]
sergio-costas: Done on all cases, not only on this one.

src/sdi-progress-dock.c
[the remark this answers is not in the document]
robert-ancell: I'd side with @3v1n0 here! The parenthesis makes it harder for me to read too.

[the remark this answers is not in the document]
sergio-costas: Removed.

tests/test-sdi-progress-window.c
sergio-costas: In fact, the only way of running both the window, dock and notifications tests in an automated fashion would be using OpenQA, because they require to check things on windows and notifications, and to do clicks over them.

Maybe the notifications test could be automated without OpenQA by emulating all the DBus notifications API, but I think that it is too much work for a piece of code that is designed to not require changes.

sergio-costas: Mmm... maybe I can play with the environment variables... I'll do a test.

sergio-costas: @3v1n0 Ok, done. If you can review the change when you have some spare time, please.

tests/test-sdi-progress-window.c
sergio-costas: The point is that the `set_progress_bar()`  callback is called several times. I can't use `g_timeout_add_once` because, in that case, the progress bar in the window would update only once, which is not what it should do.

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+  done_tasks++;
+  if (done_tasks > total_tasks)
+    done_tasks = 0;
+  return G_SOURCE_CONTINUE;
+}
+
+static void wait_for_click() {
+  loop = g_main_loop_new(NULL, FALSE);
+  g_main_loop_run(loop);
+  g_main_loop_unref(loop);
```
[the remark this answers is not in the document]
sergio-costas: `g_main_context_iteration` is, in my opinion, an ugly solution. About the flag... that would require the tests to be event based, which makes it more complex.

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+static void show_progress_window(gchar *snap_name, gchar *desktop_file) {
+  g_autoptr(GDesktopAppInfo) app_info = g_desktop_app_info_new(desktop_file);
+  g_assert_nonnull(app_info);
+  g_autofree gchar *icon = g_desktop_app_info_get_string(app_info, "Icon");
+  sdi_progress_window_begin_refresh(
+      progress_window, snap_name,
+      (gchar *)g_app_info_get_display_name(G_APP_INFO(app_info)), icon, NULL);
+}
+
+static int count_progress_childs() {
```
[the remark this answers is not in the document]
sergio-costas: GDesktopAppInfo inherits from GAppInfo, so it isn't needed. But I'll add a comment.

tests/meson.build
```diff
@@ -31,4 +31,25 @@ test_sdi_notices_monitor = executable(
...
+  '../src/sdi-progress-window.c',
+  '../src/sdi-refresh-dialog.c',
+  resources,
+  dependencies: [gtk_dep, snapd_glib_dep, gio_dep],
+  c_args: ['-DDEBUG_TESTS']
+)
+
 subdir('data')
+
+test('Tests', test_executable)
```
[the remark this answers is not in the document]
sergio-costas: The problem is that we need to access the true GSettings, to get the applications in the dock.

src/sdi-progress-window.c
```diff
@@ -0,0 +1,203 @@
...
+static void remove_dialog_from_main_window(SdiProgressWindow *self,
+                                           SdiRefreshDialog *dialog) {
+  if (dialog == NULL)
+    return;
+
+  g_hash_table_remove(self->dialogs, sdi_refresh_dialog_get_app_name(dialog));
+
+  if (!check_parent_contains_child(GTK_WIDGET(self->refresh_bar_container),
+                                   GTK_WIDGET(dialog)))
+    return;
```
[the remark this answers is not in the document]
sergio-costas: Done.

src/sdi-progress-window.c
```diff
@@ -0,0 +1,203 @@
...
+ * This callback should be connected to the `begin-refresh` signal from a
+ * #sdi_refresh_monitor object. It will create a new window if required, and
+ * insert into it a new #sdi_refresh_dialog with the snap name, snap icon and
+ * progress bar.
+ */
+void sdi_progress_window_begin_refresh(SdiProgressWindow *self,
+                                       gchar *snap_name, gchar *visible_name,
+                                       gchar *icon, gpointer data) {
+  g_autoptr(SdiRefreshDialog) dialog =
+      g_object_ref_sink(sdi_refresh_dialog_new(snap_name, visible_name));
```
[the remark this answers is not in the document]
sergio-costas: Because I add the dialog into the hash table too, so I need to add an extra reference. This forces me to sink it first.

src/sdi-progress-window.h
```diff
@@ -0,0 +1,53 @@
...
+                                       gchar *icon, gpointer data);
+
+void sdi_progress_window_update_progress(SdiProgressWindow *self,
+                                         gchar *snap_name, GStrv desktop_files,
+                                         gchar *task_description,
+                                         guint done_tasks, guint total_tasks,
+                                         gboolean task_done, gpointer data);
+
+void sdi_progress_window_end_refresh(SdiProgressWindow *self, gchar *snap_name,
+                                     gpointer data);
```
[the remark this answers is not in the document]
sergio-costas: Fixed.

src/sdi-refresh-dialog.c
```diff
@@ -155,7 +155,7 @@ void sdi_refresh_dialog_set_percentage_progress(SdiRefreshDialog *self,
   gtk_progress_bar_set_fraction(self->progress_bar, percent);
   gtk_progress_bar_set_show_text(self->progress_bar, TRUE);
   if ((bar_text != NULL) && (bar_text[0] == 0)) {
-    gtk_progress_bar_set_text(self->progress_bar, NULL);
+    gtk_progress_bar_set_text(self->progress_bar, "");
```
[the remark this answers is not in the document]
sergio-costas: Yes, you are right.

src/sdi-refresh-monitor.c
```diff
@@ -312,98 +206,105 @@ static void update_inhibited_snaps(SdiRefreshMonitor *self, SnapdChange *change,
...
+      // If the Change is completed, emit the `end-refresh` signal to close
+      // any Dialog that belongs to this snap...
+      g_signal_emit_by_name(self, "end-refresh", sdi_snap_get_name(snap));
       remove_snap(self, snap);
-      SnapRefreshData *data = snap_refresh_data_new(self, NULL, snap_name);
+      // and show, if Done, a notification to inform the user that the snap
+      // has been refreshed and they can launch it again.
       if (done) {
+        // the `data` memory block is freed in the callback, so it must not
+        // use g_autoptr here.
```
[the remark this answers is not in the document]
sergio-costas: Unified the comments.

src/sdi-refresh-monitor.c
```diff
@@ -433,13 +340,21 @@ static void update_dock_snaps(SdiRefreshMonitor *self, SnapdChange *change,
       progress_task_data->done = task_done;
       if (task_done) {
         progress_task_data->done_tasks++;
+      } else if ((progress_task_data->task_description == NULL) &&
```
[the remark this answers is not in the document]
sergio-costas: You win :-P

src/sdi-refresh-monitor.c
```diff
@@ -448,14 +363,20 @@ static void update_dock_snaps(SdiRefreshMonitor *self, SnapdChange *change,
 
 static gboolean cancelled_change_status(const gchar *status) {
   return g_str_equal(status, "Undoing") || g_str_equal(status, "Undone") ||
-         g_str_equal(status, "Undo") || g_str_equal(status, "Error");
+         g_str_equal(status, "Undo") || g_str_equal(status, "Error") ||
+         g_str_equal(status, "Abort");
```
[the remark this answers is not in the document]
sergio-costas: Mmm... yes.

tests/meson.build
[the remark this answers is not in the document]
sergio-costas: (it is used only as read)

tests/test-sdi-progress-window.c
```diff
@@ -0,0 +1,318 @@
...
+     "The first progress bar must disappear, remaining only the second "
+     "progress bar. Neither a new window, nor a new progress bar, should "
+     "appear (check below "
+     "this window just in case).\n",
+     "Confirm that is correct.", test_dual_progress_bar3},
+    {8, "/progress_window/test_dual_progress_bar_4",
+     "The window must disappear and no new window should appear (check below "
+     "this window just in case).\n",
+     "Confirm that is correct.", test_dual_progress_bar4},
+    {-1, NULL, NULL, NULL, NULL}};
```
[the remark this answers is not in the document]
sergio-costas: Done.

tests/test-sdi-progress-dock.c
```diff
@@ -0,0 +1,164 @@
...
+}
+
+int main(int argc, char **argv) {
+  g_test_init(&argc, &argv, NULL);
+
+  g_autoptr(GApplication) app = G_APPLICATION(gtk_application_new(
+      "io.snapcraft.SdiProgressDockTest", G_APPLICATION_DEFAULT_FLAGS));
+  g_signal_connect(app, "startup", (GCallback)do_startup, NULL);
+  g_signal_connect(app, "activate", (GCallback)do_activate, NULL);
+  g_application_run(app, argc, argv);
```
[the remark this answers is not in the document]
sergio-costas: I've been checking several examples in GIO, and they use several different approaches. One is to do a class derived from GApplication and do the tests inside the `activate` or `startup` methods, which is quite complex, in my opinion. In one of this kind, it has a big `switch/case` function implementing a state machine, which makes it even more complex and error-prone. In others, they connect the signals exactly like I'm doing and wait for specific events; the only difference between my code and theirs is that they use `g_main_context_iteration()` while I create a new `mainloop()`.

src/sdi-refresh-monitor.c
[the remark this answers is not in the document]
3v1n0: Yeah, it's just more clear how the hownership goes. However you can avoid using the autoptr if you're stealing, but steal is a good practice to show how memory ownership is passed around and avoids messing with the same pointer twice in a func.

tests/test-sdi-progress-window.c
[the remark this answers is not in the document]
3v1n0: `g_main_context_iteration` gives you better control without messing with the loop that is more automated, but both are fines here.

tests/meson.build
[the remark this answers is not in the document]
3v1n0: Well, I feel it's just better to use memory backend *and* before the test adding such application to the dock settings

src/sdi-progress-dock.c
```diff
@@ -0,0 +1,97 @@
...
+ * #sdi_refresh_monitor object. It will receive the total number of tasks and
+ * how many have been done, and if the task has been completed, and with that
+ * will update the progress bars in the dock.
+ */
+void sdi_progress_dock_update_progress(SdiProgressDock *self, gchar *snap_name,
+                                       GStrv desktop_files,
+                                       gchar *task_description,
+                                       guint done_tasks, guint total_tasks,
+                                       gboolean task_done) {
+  if (desktop_files != NULL && total_tasks != 0) {
```
3v1n0, line 48: Nit, just return early in this case:

```suggestion
  if (desktop_files == NULL || !total_tasks)
     return;
```

src/sdi-progress-dock.c
```diff
@@ -0,0 +1,97 @@
...
+          g_variant_new_double(done_tasks / ((gfloat)total_tasks)));
+      g_variant_builder_add(builder, "{sv}", "progress-visible",
+                            g_variant_new_boolean(!task_done));
+      g_variant_builder_add(builder, "{sv}", "updating",
+                            g_variant_new_boolean(!task_done));
+
+      g_autoptr(GVariant) values =
+          g_variant_ref_sink(g_variant_builder_end(builder));
+      unity_com_canonical_unity_launcher_entry_emit_update(
+          self->unity_manager, *desktop_file, values);
```
3v1n0, line 65: Being a variant floating you pass to dbus, it should be consumed by it already... So I'd just do:

```suggestion
      unity_com_canonical_unity_launcher_entry_emit_update(
          self->unity_manager, *desktop_file, g_variant_builder_end(builder));
```

src/sdi-progress-dock.c
```diff
@@ -0,0 +1,97 @@
...
+  GObjectClass *gobject_class = G_OBJECT_CLASS(klass);
+
+  gobject_class->dispose = sdi_progress_dock_dispose;
+}
+
+static void sdi_progress_dock_init(SdiProgressDock *self) {}
+
+SdiProgressDock *sdi_progress_dock_new(GApplication *application) {
+  SdiProgressDock *self = g_object_new(SDI_TYPE_PROGRESS_DOCK, NULL);
+  self->application = g_object_ref(application);
```
3v1n0, line 89: Nit, but it's nice:

```suggestion
  g_set_object(&self->application, application);
```

src/sdi-progress-dock.c
```diff
@@ -0,0 +1,97 @@
...
+  g_autofree gchar *unity_object =
+      g_strdup_printf("/com/canonical/unity/launcherentry/%d", getpid());
+  self->unity_manager = unity_com_canonical_unity_launcher_entry_skeleton_new();
+  g_dbus_interface_skeleton_export(
+      G_DBUS_INTERFACE_SKELETON(self->unity_manager),
+      g_application_get_dbus_connection(application), unity_object, NULL);
```
3v1n0, line 95: I feel we should handle this error, at least by logging it.

You can also return accordingly, so that caller is aware of the failure.

If you want the clean solution instead, you should implement this as a `GInitable`.

src/sdi-progress-dock.c
sergio-costas: Done, although being `total_tasks` a counter, I prefer to keep the comparison with zero.

src/sdi-progress-dock.h
```diff
@@ -0,0 +1,37 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#pragma once
+
+#include <gio/gio.h>
```
3v1n0, line 20: This can be probably just gobject no?

src/sdi-progress-dock.c
sergio-costas: Done.

src/sdi-snap.c
```diff
@@ -17,76 +17,38 @@
...
+  //* user to close the instance to continue the refresh. Non-inhibited snaps
+  //* must NOT show a progress bar.
   gboolean inhibited;
 
   //* If a dialog is destroyed for whatever reason (for example, because the top
-  // window containing it is closed) this is set to TRUE to avoid creating again
-  // a progress bar for this snap in the next loop.
+  //* window containing it is closed) this is set to TRUE to avoid creating
+  // again
+  //* a progress bar for this snap in the next loop.
```
3v1n0, line 40: Using both `//` and `*`?

src/sdi-progress-dock.c
3v1n0: Yeah, fair enough... I was undecided too :)

src/sdi-snap.c
sergio-costas: Mmm... I suspect that it was the formater. Fixed.

src/sdi-progress-dock.h
3v1n0: Oh, sorry... I missed the `GApplication` ptr. That's right!

src/sdi-progress-dock.c
sergio-costas: I added a login warning. Check it when you have some spare time, please.

src/sdi-refresh-monitor.c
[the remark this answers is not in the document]
sergio-costas: Again missed these inside the folded zone... Give me a minute...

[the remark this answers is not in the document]
sergio-costas: Done.

tests/meson.build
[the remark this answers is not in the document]
sergio-costas: But no application is added to the dock... I just read which applications are currently in the dock.

tests/test-sdi-progress-window.c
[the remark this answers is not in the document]
sergio-costas: I've been thinking and you are right... `g_main_context_iteration()` is better. Changed.

tests/test-sdi-progress-dock.c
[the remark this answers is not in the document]
sergio-costas: @3v1n0 Ok, I did a complete rewrite of the dock test. Now it is fully automated, so it runs in the CI, and I think that it does it as you say.

tests/meson.build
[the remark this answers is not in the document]
sergio-costas: Now no test uses GSettings.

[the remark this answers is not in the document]
sergio-costas: Well... not really: `test-snapd-desktop-integration` does use it, but specifically uses its own files by using `GSETTINGS_BACKEND=keyfile` and a custoM `XDG_CONFIG_HOME`, so there is no problem.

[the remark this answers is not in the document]
sergio-costas: Than test can't use the memory backend because it launches the full daemon and checks what things it does store in the gsettings.

.github/workflows/unitary-test.yaml
```diff
@@ -41,6 +42,7 @@ jobs:
 
           CHECK_COVERAGE_DATES () {
             for FILE in "${@:3}"; do
+              echo Comparing $FILE $(COMMIT_TIMESTAMP $FILE) and coverage/$1 $(COMMIT_TIMESTAMP coverage/$1)
```
3v1n0, line 45:
```suggestion
              echo "Comparing $FILE $(COMMIT_TIMESTAMP $FILE) and coverage/$1 $(COMMIT_TIMESTAMP coverage/$1)"
```

tests/test-sdi-progress-dock.c
```diff
@@ -1,160 +1,147 @@
...
+  g_assert_false(timeout_happened);
+  timeout_happened = TRUE;
 }
 
-static void wait_for_click() {
+static void wait_for_events(guint timeout, guint changes) {
   GMainContext *context = g_main_context_default();
-  clicked_on_wait = FALSE;
+  timeout_happened = FALSE;
+  guint id = g_timeout_add_once(timeout, (GSourceOnceFunc)timeoutCB, NULL);
```
3v1n0, line 36: I'd like it better if you'd pass the value:

```c
g_timeout_add_once(timeout, (GSourceOnceFunc)timeoutCB, &timeout_happened);
/* then... */
static void timeoutCB(gboolean *data) {
  gboolean *timeout_happened = data;
  g_assert_nonnull(timeout_happened);
  g_assert_false(*timeout_happened);
  *timeout_happened = TRUE;
}
```
So we avoid static global values...

3v1n0:
Looks good now from my POV, some minor nits.

Also consider squashing the commits so that we don't have many fixup ones

tests/test-sdi-progress-dock.c
sergio-costas: Mmm... I like it!!!

sergio-costas: Done. And also rebased everything.

tests/test-sdi-progress-dock.c
```diff
@@ -0,0 +1,154 @@
...
+static void wait_for_events(guint timeout, guint changes) {
+  GMainContext *context = g_main_context_default();
+  guint timeout_id =
+      g_timeout_add_once(timeout, (GSourceOnceFunc)timeoutCB, NULL);
+  do {
+    g_main_context_iteration(context, TRUE);
+  } while ((timeout_id != 0) && (current_changes & changes) != changes);
+  if (timeout_id != 0) {
+    g_clear_handle_id(&timeout_id, g_source_remove);
+  }
```
3v1n0, line 41: No need for the if, that's the point of the `g_clear_*` thingy :)

```suggestion
  g_clear_handle_id(&timeout_id, g_source_remove);
```

tests/test-sdi-progress-dock.c
```diff
@@ -0,0 +1,154 @@
...
+  *timeout_id = 0;
+}
+
+static void wait_for_events(guint timeout, guint changes) {
+  GMainContext *context = g_main_context_default();
+  guint timeout_id =
+      g_timeout_add_once(timeout, (GSourceOnceFunc)timeoutCB, NULL);
+  do {
+    g_main_context_iteration(context, TRUE);
+  } while ((timeout_id != 0) && (current_changes & changes) != changes);
```
3v1n0, line 38:
```suggestion
  } while (timeout_id != 0 && (current_changes & changes) != changes);
```

tests/test-sdi-progress-dock.c
sergio-costas: Agh! Of course! 😅

3v1n0 (approved):

sergio-costas:
Thank you for taking so much time and work for reviewing it.

PR #100, opened 2024-09-26
Move false helpers where they are used

sergio-costas:
Several helper functions were used only in a single .c file, thus being "fake helpers". This PR moves them into their corresponding file, makes them static to avoid exporting them, and removes the `sdi_` prefix.


robert-ancell (approved):
LGTM

PR #101, opened 2024-09-26
Clean exit (#98)

sergio-costas:
* Clean exit

This patch manages the INT and TERM signals to cleanly quit the main loop and free all the used resources.

* Clear objects in shutdown event

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -37,7 +37,6 @@ static Login1Manager *login_manager = NULL;
 static SnapdClient *client = NULL;
 static SdiThemeMonitor *theme_monitor = NULL;
 static SdiRefreshMonitor *refresh_monitor = NULL;
-static gboolean running = FALSE;
 
 static gchar *snapd_socket_path = NULL;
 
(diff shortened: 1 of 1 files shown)


sergio-costas:
@robert-ancell I accidentally set as destination branch an incorrect one. The "clean exit" patch should have been merged into main. Any problem with me merging it?

src/main.c
```diff
@@ -150,7 +151,8 @@ static void do_startup(GObject *object, gpointer data) {
 
 static void do_activate(GObject *object, gpointer data) {
   // because, by default, there are no windows, so the application would quit
-  g_application_hold(G_APPLICATION(app));
+  g_application_hold(G_APPLICATION(object));
+  running = TRUE;
```
3v1n0, line 155: Can't you just hold the app by default and avoid the `running` bit?

sergio-costas: The problem is that pressing Ctrl-C seems to send both SIGINT and SIGTERM signals, so g_application_release() is called twice.

sergio-costas: (part of the problem is due to the trick used to avoid problems with text-only sessions)

sergio-costas: Ok, I found that I can call g_application_quit several times, so that fixes the problem.

robert-ancell (approved):
LGTM, but perhaps wait for @3v1n0 to finish his review.

3v1n0:
Thanks for asking, but all good from my side too! :)

sergio-costas:
@robert-ancell @3v1n0 Thanks! Merging.

PR #102, opened 2024-09-27
Refactor notifications and add unitary tests

sergio-costas:
This PR does some refactoring on the `sdi-notify` object to allow to be connected through signals to the `sdi-refresh-monitor` object (mainly replace the GSList with the list of applications that have available refreshes, with a GListModel, a GObject with the same functionality, which is easier to send through signals). It also moves the little function that checks if a refresh is or not forced, from the `notify` object into the `monitor` object, thus ensuring that the `notify` object is just as "dumb" as possible about the refresh process.

This change not only decouples more the code between the refresh monitor and the notifications object; it also allows to add ten unitary tests for the notifications code (although they can't be run automatically because they require to connect to the desktop notifications daemon; maybe with OpenQA, or using an emulation of the notification daemon...) to ensure that the `sdi-notify` object works as expected.

### DIFF
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -23,6 +23,7 @@
 #include <glib/gi18n.h>
 #include <libnotify/notify.h>
 
+#include "io.snapcraft.PrivilegedDesktopLauncher.h"
 #include "sdi-forced-refresh-time-constants.h"
 #include "sdi-helpers.h"
 #include "sdi-notify.h"
@@ -36,13 +37,6 @@ struct _SdiNotify {
 };
 
 G_DEFINE_TYPE(SdiNotify, sdi_notify, G_TYPE_OBJECT)
-
-static GTimeSpan get_remaining_time_in_seconds(SnapdSnap *snap) {
-  GDateTime *proceed_time = snapd_snap_get_proceed_time(snap);
-  g_autoptr(GDateTime) now = g_date_time_new_now_local();
-  GTimeSpan difference = g_date_time_difference(proceed_time, now) / 1000000;
-  return difference;
-}
 
 static gboolean launch_desktop(GApplication *app, const gchar *desktop_file) {
   g_autofree gchar *full_desktop_path = NULL;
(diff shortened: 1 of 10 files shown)


src/sdi-notify.c
```diff
@@ -294,8 +298,8 @@ static void show_pending_update_notification(SdiNotify *self,
   if (allow_to_ignore) {
     g_autoptr(GVariantBuilder) builder =
         g_variant_builder_new(G_VARIANT_TYPE("as"));
-    for (; snaps != NULL; snaps = snaps->next) {
-      SnapdSnap *snap = (SnapdSnap *)snaps->data;
+    for (int i = 0; i != g_list_model_get_n_items(snaps); i++) {
```
robert-ancell, line 301: nit: Conventionally this would be `<` instead of `!=`

src/sdi-refresh-monitor.c
```diff
@@ -512,10 +540,11 @@ static void manage_refresh_inhibit(SnapdClient *source, GAsyncResult *res,
     return;
   // Check if there's at least one snap not marked as "ignore"
   gboolean show_grouped_notification = FALSE;
-  g_autoptr(GSList) snap_list = NULL;
+  g_autoptr(GListStore) snap_list = g_list_store_new(SNAPD_TYPE_SNAP);
   for (guint i = 0; i < snaps->len; i++) {
     SnapdSnap *snap = snaps->pdata[i];
     const gchar *name = snapd_snap_get_name(snap);
+    g_debug("Received notification for inhibited snap %s", name);
```
robert-ancell, line 547: Is this suppose to remain? (I think so, just double checking)

src/sdi-refresh-monitor.c
```diff
@@ -645,65 +675,37 @@ void sdi_refresh_monitor_init(SdiRefreshMonitor *self) {
   configure_snapd_monitor(self);
 }
 
-static void ignore_snap_cb(GObject *obj, const gchar *snap_name,
-                           SdiRefreshMonitor *self) {
+/**
+ * This CB must be called whenever the user presses the "Don't remind me
```
robert-ancell, line 679: CB => callback? Seems unnecessary to abbreviate.

robert-ancell (approved):
Some minor questions but otherwise looks good!

src/sdi-notify.c
sergio-costas: Yes, you are right. Fixed.

src/sdi-refresh-monitor.c
sergio-costas: Yes. I want to add some extra debug messages, and maybe enable them when launched as a service to simplify fixing bugs (that way we can ask for a journalctl dump).

src/sdi-refresh-monitor.c
sergio-costas: Absolutely right. Fixing it...

PR #105, opened 2024-09-30
Fix memory leak

sergio-costas:
The call to get_name_from_snap() returns a new block of memory, which must be freed after being used. Instead, in sdi-notify, the function was being called directly in the parameters list of another call, thus the block was left around.

This patch fixes it.


sergio-costas:
I'm not sure which style will convince you more: either this one, or the one in #106.

sergio-costas:
Closing after suggestion from Robert.

PR #106, opened 2024-10-01
Fix memory leak alternate version

sergio-costas:
This is an alternative version to #105

### DIFF
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -371,22 +371,22 @@ static gchar *get_name_from_snap(SnapdSnap *snap) {
   return g_strdup(name);
 }
 
-static gchar *build_body_message_for_two_refreshes(GSList *snaps) {
-  g_autofree gchar *snap_name0 = get_name_from_snap((SnapdSnap *)snaps->data);
-  g_autofree gchar *snap_name1 =
-      get_name_from_snap((SnapdSnap *)snaps->next->data);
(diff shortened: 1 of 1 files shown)


src/sdi-notify.c
```diff
@@ -371,6 +371,28 @@ static gchar *get_name_from_snap(SnapdSnap *snap) {
...
+  g_autofree gchar *snap_name0 = get_name_from_snap((SnapdSnap *)snaps->data);
+  g_autofree gchar *snap_name1 =
+      get_name_from_snap((SnapdSnap *)snaps->next->data);
+  /// TRANSLATORS: This message is used when there are two pending
+  /// refreshes.
+  return g_strdup_printf(_("%s and %s will update when you quit them."),
+                         snap_name0, snap_name1);
+}
+
+static gchar *build_body_message_for_three_refreshes(GSList *snaps) {
```
3v1n0, line 384: I think it would be nicer if this function takes two `SnapdSnap` parameters instead so that you don't have to access to the `GList` internals (as you control the safety of that here but only in the caller function).

src/sdi-notify.c
```diff
@@ -371,6 +371,28 @@ static gchar *get_name_from_snap(SnapdSnap *snap) {
   return g_strdup(name);
 }
 
+static gchar *build_body_message_for_two_refreshes(GSList *snaps) {
```
3v1n0, line 374: Same here...

sergio-costas: Done too.

PR #108, opened 2024-10-02
Update to Core24

sergio-costas:
Tested with Core24, and everything works fine: both the dock progress bar and the dialogs with the current progress.


sergio-costas:
Also, the test for themes changes does pass.

seb128 (approved):
Thanks!

PR #109, opened 2024-10-03
Revert "Update to Core24 (#108)"

seb128:
This reverts commit 0bd45e0db93bd9f5463df5755761ee61efd03d3d.

We decided to keep core22 snaps for Oracular


sergio-costas (approved):
Since finally there will be Core22 snaps in 24.10, it makes sense to not rush this change.

PR #110, opened 2024-10-03
Add instructions on how to report security issues

seb128:
Quoting SEC0026 - SSDLC

> SECURITY.md (V1.1)
> Any public repository must also include a SECURITY.md file in the root directory, which educates users/contributors on
> how to report a security concern.

The content of the file is based on the template referenced in the spec and what other Canonical repositories are already using


sergio-costas (approved):
The text is simple and direct to the point.

PR #113, opened 2024-10-10
Run unitary tests

sergio-costas:
This PR adds a new job in the `tests` for running the `themes change` unitary test. In the future, more unitary tests will be run here.


sergio-costas:
It requires a running graphical environment :-(

PR #118, opened 2024-12-09
Add meson_version to meson.build

sergio-costas:
Adds a check to ensure that at least meson 0.64.0 or later is used, since it is required for fs.copyfile.


seb128 (approved):
Thanks!

PR #121, opened 2024-12-11
Using format-source.sh for checking the format

sergio-costas:
This PR adds the format-source.sh script to simplify checking the source format.


3v1n0:
Can we maybe move this to https://github.com/canonical/desktop-engineering/ so that we can reuse them in different projects?

Potentially adding an input to choose a different clang configuration, for potential configs that are not using `.clang-format` files?

sergio-costas:
@3v1n0 Sent: https://github.com/canonical/desktop-engineering/pull/56

sergio-costas:
Updated the script with the latest changes in desktop-engineering

sergio-costas:
@aleasto @3v1n0 So... what do you think...?

PR #122, opened 2024-12-11
Coverage support for unitary tests

sergio-costas:
This PR adds support for COBERTURA.XML generation. The CI will generate the .xml file and upload it as an artifact called `tics-coverage-reports`.


.github/workflows/unitary-test.yaml
```diff
@@ -29,3 +29,31 @@ jobs:
...
+              if [[ $(COMMIT_TIMESTAMP $FILE) > $(COMMIT_TIMESTAMP coverage/$1) ]]; then
+                echo $FILE has been modified. You must run *$2* to update coverage
+                exit 1
+              fi
+            done
+          }
+
+          # Check if the prebuilt COBERTURA.XML files are up-to-date
+          CHECK_COVERAGE_DATES cobertura-notify.xml run-test-sdi-notify.sh tests/test-sdi-notify.c src/sdi-notify.c src/sdi-helpers.c
+          CHECK_COVERAGE_DATES cobertura-theme-change.xml run-test-theme-change.sh tests/test-snapd-desktop-integration.c src/sdi-theme-monitor.c src/sdi-theme-monitor.h src/main.c
```
d-loose, line 49: As discussed via chat, I'd try to automate at least some parts of this in the future (e.g. by using something like `xvfb-run` where possible) to not have to manually update the coverage files, but this is a very nice workaround!

sergio-costas: @d-loose Updated the tests to extract the coverage data from the test that I moved into CI. It should be ready.

d-loose (approved):
LGTM :+1:

PR #125, opened 2025-01-09
Fix tiobe errors

sergio-costas:
This patch fixes some of the Tiobe errors.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -124,7 +124,7 @@ static void do_shutdown(GObject *object, gpointer data) {
 
 static int global_retval = 0;
 
-void sighandler(int v) {
+static void sighandler(int v) {
   global_retval = 128 + v; // exit value is usually 128 + signal_id
 }
 
--- a/src/main.c
+++ b/src/main.c
@@ -55,7 +55,7 @@ static GOptionEntry entries[] = {{"snapd-socket-path", 0, G_OPTION_FLAG_NONE,
 static void do_startup(GObject *object, gpointer data) {
   sdi_snapd_client_factory_set_custom_path(snapd_socket_path);
 
-  refresh_monitor = sdi_refresh_monitor_new(G_APPLICATION(object));
+  refresh_monitor = sdi_refresh_monitor_new();
 
   notify_manager = sdi_notify_new(G_APPLICATION(object));
   g_signal_connect_object(refresh_monitor, "notify-pending-refresh",
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -23,7 +23,6 @@
 #include <gio/gdesktopappinfo.h>
 #include <glib/gi18n.h>
 #include <libnotify/notify.h>
-#include <stdbool.h>
 
 #include "io.snapcraft.PrivilegedDesktopLauncher.h"
 #include "sdi-helpers.h"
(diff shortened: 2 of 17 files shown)


src/sdi-notify.c
```diff
@@ -38,7 +38,7 @@ struct _SdiNotify {
 
 G_DEFINE_TYPE(SdiNotify, sdi_notify, G_TYPE_OBJECT)
 
-static gboolean launch_desktop(GApplication *app, const gchar *desktop_file) {
+static bool launch_desktop(GApplication *app, const gchar *desktop_file) {
```
d-loose, line 41: To me this looks like the linter used by tiobe isn't aware of glib conventions. But I don't have much experience with glib, so I'd suggest you ask someone more knowledgeable than me to review this :)

sergio-costas: Yes, that's a good part of the problem: glib predates stdbool, so it uses an `int` as a boolean, FALSE is defined as `0` and TRUE is `!!FALSE`, which makes a "funny" mix. That's why I changed this one, but not others, where the gboolean is a must for some reason.

sergio-costas: But you are right... it makes no sense to fix something that is fine... I'll revert the gboolean -> bool changes

sergio-costas: Done.

src/main.c
```diff
@@ -55,7 +55,7 @@ static GOptionEntry entries[] = {{"snapd-socket-path", 0, G_OPTION_FLAG_NONE,
 static void do_startup(GObject *object, gpointer data) {
   sdi_snapd_client_factory_set_custom_path(snapd_socket_path);
 
-  refresh_monitor = sdi_refresh_monitor_new(G_APPLICATION(object));
+  refresh_monitor = sdi_refresh_monitor_new();
```
d-loose, line 58: This also doesn't look right to me

src/sdi-notify.c
d-loose: There are a couple of `TRUE`->`true` and `FALSE`->`false` changes, that probably need to be reverted as well.

src/main.c
sergio-costas: That's because the application isn't used in sdi_refresh_monitor_new(). It was used some PRs before, but in a refactor it wasn't needed, but accidentally forgot to remove it from the parameters, so TIOBE complains that the parameter isn't used in that function.

src/sdi-notify.c
sergio-costas: In fact, those two cases are errors: they should be as they are now, because they are gboolean, so they should use TRUE and FALSE.

src/sdi-refresh-monitor.c
```diff
@@ -255,7 +260,7 @@ static void process_inhibited_snaps(SdiRefreshMonitor *self,
                               NULL);
       } else {
         // If we have snap data, we can use "pretty names" and icons
-        const gchar *snap_name = snapd_snap_get_name(client_snap);
+        const gchar *snap_name2 = snapd_snap_get_name(client_snap);
```
d-loose, line 263: I suppose tiobe complained about shadowing the existing `snap_name` variable here? In any case, I'd use a more expressive name - it's not clear to me how `snap_name` and `snap_name2` are different, since you're obtaining `client_snap` from `snap_name` via `snapd_client_get_snap_sync`, and then use `client_snap` to get `snap_name2` using `snapd_snap_get_name` (doesn't this just return the name you already have?_

sergio-costas: You are right... I think that the name is the same, but I prefer to wait to have all the unitary tests in place before modifying this. Reverted to the original code.

d-loose (approved):

PR #127, opened 2025-01-09
Add TICS workflow for CI/CD

sergio-costas:
This patch uses a different CI/CD workflow to generate the coverage data and upload it to tiobe.


d-loose (approved):
Had a look at the documentation and the examples in other repositories - this matches what I've seen there, so LGTM :+1:

PR #131, opened 2025-01-21
Move the pre-built coverage files into coverage_base

sergio-costas:
The first partically successful test shown several errors saying that there were duplicated entries for coverage; specifically, those were four times in the data. This matches with the fact that there are four coberturaXXX.xml files inside the coverage folder: two pre-computed with the manual tests, another created during the CI/CD pass for the rest of the code, and the final one which merges those three into a single file.

This patch moves the pre-computed and the temporal cobertura files into coverage_base folder, and leaves only the final cobertura.xml file inside the coverage folder. It should fix that bug.


d-loose (approved):
Thanks for the explanation, that makes sense to me :+1:

PR #132, opened 2025-01-21
Add tests coverage

sergio-costas:
This patch adds coverage to the tests files themselves.

The key part is in the gcovr.cfg file, where the tests folder is added. Of course, the pre-built coverage files had to be rebuilt.


sergio-costas:
I found that there is a way of marking functions as "dead code", so they can avoid this kind of errors. I'll add them to the test files.

d-loose:
I'm a bit confused about this - why would you need coverage data for the tests?

sergio-costas:
@d-loose To avoid an error shown during the execution of the TIOBE upload:

imagen

This is not really a big problem according to the TIOBE people. They said in a mail:

> For code coverage coverage, TiCS always outputs “error” when it does not find a match.
> I know it is not the best info message, and that has been passed to our product team some time ago.
> To see the test code, please click “Filter-> Code type”.

So it's just to ensure that there are no odd messages.

sergio-costas:
What do you think, @seb128  ?

sergio-costas:
@seb128 @d-loose Removed those pesky commits :-)

d-loose (approved):

PR #139, opened 2025-03-26
Add support for snapd abstract socket

sergio-costas:
Snapd has changed to an abstract socket when used from inside a snap, so a change was required in snapd-glib. Now, the library automagically detects if it is inside a snap and chooses, in that case, the new abstract socket if it exists, or the old, normal socket if the abstract doesn't exist, so the current code to set the socket if we are inside a snap is not only not needed anymore, but it will stop working in the next snapd version, so it must be removed and let snapd-glib to choose the right socket.


sergio-costas:
Thanks!

aleasto:
Does this need a version check at runtime, or even at build time?

Or because we're in a snap we can tightly couple the library so that it's a non-problem?

PR #140, opened 2025-04-02
Don't show notifications with negative expiration time

sergio-costas:
Sometimes, snapd notifies us with an event that has a negative time for a forced quit, when, in fact, there are no pending updates for that snap. This is a very undesirable effect, because it confuses the final user.

This patch ensures that no notification with a negative value is shown.

fix https://github.com/canonical/snapd-desktop-integration/issues/135

### DIFF
--- a/tests/test-refresh-monitor.c
+++ b/tests/test-refresh-monitor.c
@@ -978,11 +978,11 @@ test_refresh_inhibit_with_negative_value_dont_shows_notifications(void) {
   new_notice("refresh-inhibit");
   g_assert_true(wait_for_notice());
 
-  g_autoptr(ReceivedSignalData) data2 =
+  g_autoptr(ReceivedSignalData) data =
       wait_for_signal(RECEIVED_SIGNAL_NOTIFY_PENDING_REFRESH, 100);
-  g_assert_nonnull(data2);
-  g_assert_true(snap_list_contains_name(data2, "snap1"));
-  g_assert_false(snap_list_contains_name(data2, "snap2"));
+  g_assert_nonnull(data);
+  g_assert_true(snap_list_contains_name(data, "snap1"));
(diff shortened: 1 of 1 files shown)


src/sdi-refresh-monitor.c
```diff
@@ -490,6 +490,17 @@ static void manage_refresh_inhibit(SnapdClient *source, GAsyncResult *res,
     if (snap_data == NULL) {
       continue;
     }
+
+    /* Sometimes, snapd sents a notification with a negative value.
```
aleasto, line 494: nit:
```suggestion
    /* Sometimes, snapd sends a notification with a negative value.
```

tests/test-refresh-monitor.c
```diff
@@ -966,13 +966,36 @@ test_sdi_get_desktop_file_from_snap_two_valid_apps_none_right(void) {
...
+  reset_mock_snapd();
+  MockSnap *snap1 = mock_snapd_add_snap(snapd, "snap1");
+  set_snap_as_inhibited(snap1,
+                        TIME_TO_SHOW_REMAINING_TIME_BEFORE_FORCED_REFRESH + 10);
+  MockSnap *snap2 = mock_snapd_add_snap(snapd, "snap2");
+  set_snap_as_inhibited(snap2, -400);
+  new_notice("refresh-inhibit");
+  g_assert_true(wait_for_notice());
+
+  g_autoptr(ReceivedSignalData) data2 =
```
aleasto, line 981: nit:
not sure why this is called `data2`

aleasto (approved):

tests/test-refresh-monitor.c
sergio-costas: Ops... because I copied another test and deleted a check before. I'll change that.

sergio-costas: Done.

PR #141, opened 2025-04-16
Add a workaround for Ubuntu Budgie

sergio-costas:
Ubuntu Budgie has a bug in its notifications system: when the user clicks on an action (for example, "Yes" for installing new themes from the Snap Store), it sends the button action (thus, "yes" action), but after that also sends the default action. Instead, both Gnome and KDE only sends the button action.

This results in odd behavior in this system: since the default action does the same than the "yes" action, if the user presses "Yes", two install processes will run concurrently, conflicting between them and resulting in an installation error shown (due to one of the processes being unable to install the themes), while, at the same time, the themes are correctly installed (by the second process).

Also, if the user presses "No", the "no" action is just ignored and nothing is done; but the next "default" action is honored and the themes are installed; exactly the opposite of what the user wanted.

This workaround ensures that only one action is honored for each notification, and any extra action is ignored until the install notification is closed and a new notification is shown.

Fix https://github.com/canonical/snapd-desktop-integration/issues/52

### DIFF
--- a/src/sdi-theme-monitor.c
+++ b/src/sdi-theme-monitor.c
@@ -41,6 +41,7 @@ struct _SdiThemeMonitor {
 
   /* The desktop notifications */
   NotifyNotification *install_notification;
+  gboolean install_notification_answered;
   NotifyNotification *progress_notification;
 
   // Connection to snapd.
@@ -91,12 +92,29 @@ static void notification_closed_cb(NotifyNotification *notification,
                                    SdiThemeMonitor *self) {
   /* Notification has been closed: */
   g_clear_object(&self->install_notification);
+  self->install_notification_answered = FALSE;
 }
 
 static void notify_cb(NotifyNotification *notification, gchar *action,
                       gpointer user_data) {
   SdiThemeMonitor *self = user_data;
 
+  /* Some notification daemons, when the user presses an action, emit
+   * the action first, and after it, the "default" action. Although
+   * this is an incorrect behavior, we add a workaround to avoid more
+   * bug reports.
+   *
+   * Ideally, we should block the action signal here and unlock it in
+   * the "closed" callback; unfortunately, this isn't a standard
+   * GObject signal, but a specific callback in the notification,
+   * so we have to do it "manually".
+   */
+
+  if (self->install_notification_answered == TRUE) {
+    return;
+  }
+
+  self->install_notification_answered = TRUE;
   if ((strcmp(action, "yes") == 0) || (strcmp(action, "default") == 0)) {
     g_message("Installing missing theme snaps...\n");
     self->progress_notification = notify_notification_new(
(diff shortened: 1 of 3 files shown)


src/sdi-theme-monitor.c
```diff
@@ -91,12 +92,22 @@ static void notification_closed_cb(NotifyNotification *notification,
...
 
+  /* Some notification daemons, when the user presses an action, emit
+   * the action first, and after it, the "default" action. Although
+   * this is an incorrect behavior, we add a workaround to avoid more
+   * bug reports. */
+  if (self->install_notification_answered == TRUE) {
+    return;
+  }
+
+  self->install_notification_answered = TRUE;
```
3v1n0, line 110: Can't you instead disconnect the signal on first callback and connect it back when needed?

Or bock it instead?

sergio-costas: Mmm... good point...

sergio-costas: Ok, it's not possible because they aren't signals, but specific callbacks in the notification, and it seems that it's not possible to "disable" them. Only "close" is a true GObject signal. I'll add a comment explaining it.

nteodosio (approved):

sergio-costas:
> Why do we need a default action anyways?

Because in some cases, clicking on the notification should open the store, or the updated application. This is part of the Design document.

image

image

aleasto:
> > Why do we need a default action anyways?
> 
> Because in some cases, clicking on the notification should open the store, or the updated application. This is part of the Design document.
> 
> image
> 
> image

Well, why is the "default" action hooked up to snapd_client_install_themes_async then?

aleasto:
> > Well, why is the "default" action hooked up to snapd_client_install_themes_async then?
> 
> Oh, you mean "Why the default action does the same than 'yes' action"...

More like "Why does the default action do anything at all"; but yes, that.

sergio-costas:
Also, the specification ( https://specifications.freedesktop.org/notification-spec/1.3/protocol.html ) says that a client must not presume that a notifications server can send actions... I'm not sure how to interpret that, because at least the "Closed" notification is sent. If a notifications server can't show buttons, but can, at least, send the "default" action...

But maybe this can be considered a design problem... @juanruitina  should we replace the notifications with a popup? We also should keep in mind that this works as expected in Ubuntu Desktop, and that, as today, all notification servers that I know do support actions...

juanruitina:
> the "default" action should be also the one that the users would want more to do

In most cases, the "default" action should be less of an actual action and more about getting more information about what the notification is about. Most frequently by opening the app itself, perhaps in a specific view/page where the user can take action. I'd say that's a common expectation across platforms.

I don't know enough about how screen readers read notifications, happy to look into it together.

sergio-costas:
@juanruitina The point is... what should we do in that case? I mean, this is a notification case that wasn't considered in the Figma document, because it's not about snap updates, but about Gtk themes, which is something that was already implemented. This notification is triggered when snapd-desktop-integration detects that there is a Gtk theme installed in the system (and thus, only accessible to non-snapped applications), and, although there is a snap in the snap store with the same Gtk theme (thus, only accessible to snapped applications), it isn't installed in the system, so it offers the user the opportunity to automagically install it. Currently, the default action does the same than choosing "Yes". Should we do something different?

sergio-costas:
Mmm... I think that it would require modifying the snap store... let me check, anyway...

PR #142, opened 2025-04-16
Update ubuntu version in test

sergio-costas:
The ubuntu 20.04 actions runner is being deprecated, so we must update it to a more recent one.

https://github.com/actions/runner-images/issues/11101


seb128 (approved):
Thanks!

PR #147, opened 2025-05-28
Add check based on DBus well-known names

sergio-costas:
Since, in some cases, snapd-desktop-integration is launched too soon, and that triggers launching xdg-desktop-portal too soon too (because several services from mutter aren't still available in DBus, so xdg-desktop-portal presumes that they don't exist).

To really fix this, a change in snapd is required that allows to specify that a service (snapd-desktop-integration in this case) must be bound to the graphical-session target instead of the default target.

But since this will require time, in the meantime this patch should fix this problem: it waits until any of the following well-known names do appear in the bus:

    org.freedesktop.portal.Desktop
    org.freedesktop.impl.portal.desktop.*

When that happens, we can assume that the desktop is ready, and launch snapd-desktop-integration. It's an ugly hack, but it should work until full graphical-session target support is added to snapd.

Fix https://bugs.launchpad.net/xdg-desktop-portal/+bug/2042301/comments/22


sergio-costas:
@aleasto When you have some spare time, can you test that this fixes the problem with the nVidia cards, please?

src/sdi-user-session-helper.c
```diff
@@ -111,7 +119,40 @@ static void new_session(Login1Manager *manager, const gchar *session_id,
...
-void sdi_wait_for_graphical_session(void) {
+static void dbus_timeout_cb(GMainLoop *loop) {
+  dbus_timeout_id = 0;
+  g_main_loop_quit(loop);
+}
+
+static gboolean check_dbus_names_cb(GMainLoop *loop) {
+  g_auto(GStrv) names = NULL;
+  GError *error = NULL;
+  dbus_server__call_list_names_sync(dbus_server, &names, NULL, &error);
```
3v1n0, line 130: This can be quite slow, it would be nicer to have it async, although it's also true we're still blocking on this... :thinking:

3v1n0:
> To really fix this, a change in snapd is required that allows to specify that a service (snapd-desktop-integration in this case) must be bound to the graphical-session target instead of the default target.

Mh, instead of waiting for well know dbus services, I think we can just use the systemd dbus API instead to just do the same, so basically wait for the graphical-target being started?

It would be just about to connect to `org.freedesktop.systemd1`'s `/org/freedesktop/systemd1/unit/graphical_2dsession_2etarget`'s `properties-changed` and check when the state changes.

You can see the various properties by `gdbus introspect --session  --dest org.freedesktop.systemd1 --object-path /org/freedesktop/systemd1/unit/graphical_2dsession_2etarget`

Note that you may need to call `Subscribe` first (see https://github.com/systemd/systemd/pull/36413).

3v1n0:
Note that I'd likely do this in a previous command-chain IMHO, so we can just add a simple python script (`wait-for-systemd-target.py`) that starts the daemon once the target is there, so that it can be easily reverted when we can just rely on systemd from snapd side.

src/sdi-user-session-helper.c
```diff
@@ -18,7 +18,13 @@
 #include "sdi-user-session-helper.h"
 
 static Login1Manager *login_manager = NULL;
+static dbusserver *dbus_server = NULL;
 static guint idle_id = 0;
+static guint dbus_check_id = 0;
+static guint dbus_timeout_id = 0;
+static bool already_in_graphical_session = false;
+static gchar *dbus_names[] = {"org.freedesktop.portal.Desktop",
+                              "*org.freedesktop.impl.portal.desktop.", NULL};
```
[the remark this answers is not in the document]
sergio-costas: That has two problems: first, that require a specific name, so I can't wait for "any name that fulfills "org.freedesktop.impl.portal.desktop.*" (which is what I do in the second case). Also, there is the other problem of having access to that signal: this implementation uses only elements already available in snapd, without having to add neither new interfaces, nor changes to the current interfaces. The same happens with waiting for the `graphical-session` target: I already implemented it, but unfortunately there is no current interface that allows us to read that (because it requires reading the `session bus`, not the `system bus`).

sergio-costas:
> Mh, instead of waiting for well know dbus services, I think we can just use the systemd dbus API instead to just do the same, so basically wait for the graphical-target being started?

As already commented, I already tried that, but there's no current interface that allows to access the session systemd instance. https://github.com/canonical/snapd-desktop-integration/pull/146

src/sdi-user-session-helper.c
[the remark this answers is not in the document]
sergio-costas: Mainly because, if we will add a new interface or a change, it's better to already do it TheRightWay(tm).

src/sdi-user-session-helper.c
sergio-costas: That's the point: we are already blocking here, so let's not overcomplicate things.

sergio-costas:
> > As already commented, I already tried that, but there's no current interface that allows to access the session systemd instance. #146
> 
> Isn't what I pasted above exactly what we want?

Oh, you mean why I didn't use that PR. As commented, the problem is that there is no current interface that allows a snapped program to get access to the session systemd, which is the one containing the targets.

3v1n0:
I'm saying something like this as initial command-chain startup script for the snap:

```python

#!/usr/bin/env python3

import os
import shutil
import sys

from gi.repository import Gio, GLib

BUS_NAME = "org.freedesktop.systemd1"
MANAGER_PATH = "/org/freedesktop/systemd1"
MANAGER_INTERFACE = "org.freedesktop.systemd1.Manager"
UNIT_INTERFACE = "org.freedesktop.systemd1.Unit"
ACTIVE_STATE = "ActiveState"

GRAPHICAL_SESSION = "graphical-session.target"

TARGET_STATE = "active"

def on_properties_changed(proxy, changed_properties, invalidated_properties, loop):
    active_state = changed_properties.lookup_value(ACTIVE_STATE, None)
    if not active_state:
        return

    if check_active_state(active_state.unpack()):
        loop.quit()

def check_active_state(active_state):
    print(f"ActiveState changed: {active_state}")
    return active_state == TARGET_STATE

def wait_for_active_state(unit_name):
    session_bus = Gio.bus_get_sync(Gio.BusType.SESSION, None)

    unit_path = f"/org/freedesktop/systemd1/unit/{unit_name.replace('-', '_2d').replace('.', '_2e')}"

    unit_proxy = Gio.DBusProxy.new_sync(
        session_bus,
        Gio.DBusProxyFlags.NONE,
        None,
        BUS_NAME,
        unit_path,
        UNIT_INTERFACE,
        None
    )

    loop = GLib.MainLoop()
    unit_proxy.connect("g-properties-changed", on_properties_changed, loop)

    initial_state = unit_proxy.get_cached_property(ACTIVE_STATE).unpack()
    if check_active_state(initial_state):
        return

    print(f"Monitoring ActiveState changes for {unit_name}. Press Ctrl+C to exit.")
    try:
        loop.run()
    except KeyboardInterrupt:
        loop.quit()

def main():
    session_bus = Gio.bus_get_sync(Gio.BusType.SESSION, None)

    manager_proxy = Gio.DBusProxy.new_sync(
        session_bus,
        Gio.DBusProxyFlags.NONE,
        None,
        BUS_NAME,
        MANAGER_PATH,
        MANAGER_INTERFACE,
        None
    )

    manager_proxy.call_sync(
        "Subscribe",
        None,
        Gio.DBusCallFlags.NONE,
        -1,
        None
    )

    wait_for_active_state(GRAPHICAL_SESSION)

    if len(sys.argv) > 1:
        os.execv(shutil.which(sys.argv[1]), [shutil.which(sys.argv[1])] + sys.argv[2:])

if __name__ == "__main__":
    main()
```

3v1n0:
Note that the session bus should be already set at that time, but if you want to test this manually you can just use `DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/$(id -u)/bus`.

3v1n0:
> The problem with that is, again, that there is no interface in snapd that allows to access those DBus objects/interfaces/methods/properties, so putting that in the snap just will fail along with a REJECTED entry in the system log.

Ohhhh... Sorry, I was not thinking on confinement... -_-

sergio-costas:
> Ohhhh... Sorry, I was not thinking on confinement... -_-

Neither did I when I wrote the first patch 😅

3v1n0:
Anyways, I'd still do the same in a launcher script, more than in the actual binary

sergio-costas:
Yes, with the new system it makes sense, because now the `gtk_init()` call is done after checking the session. Before it was done first, and if it failed, it waited for a session.

3v1n0:
Yeah, it's also not the daemon duty IMHO, so I'd just ensure that when the daemon starts everything is already all set

sergio-costas:
I'll add some documentation in the source code.

snap/snapcraft.yaml
```diff
@@ -23,19 +23,21 @@ plugs:
...
+    restart-condition: always
+    restart-delay: 1s
     plugs:
       - snap-themes-control
       - login-session-observe
       - snap
       - snap-refresh-observe
       - unity7
+      - x11
+      - wayland
```
3v1n0, line 40: Aren't these part of the gnome extension anyways?

3v1n0 (commented):
I still feel that it would be more a python job this one, but indeed having native code is always nice

However, the approach is fine given the constraints we have, but there are some cleanups to do.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,247 @@
...
+  if (got_session_list) {
+    /* check if there is already a graphical session opened for us, in which
+     * case we must just exit and let systemd to relaunch us, because it means
+     * that we run too early and the desktop wasn't still ready.
+     */
+    for (int i = 0; i < g_variant_n_children(sessions); i++) {
+      GVariant *session = g_variant_get_child_value(sessions, i);
+      if (session == NULL) {
+        continue;
+      }
```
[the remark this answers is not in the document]
sergio-costas: No, it isn't. If the session is NULL, I want to continue to the next FOR iteration, while g_assert will return (remember that a system can have several users at the same time in a multiseat environment).

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,247 @@
...
+  if (user == NULL) {
+    g_message("Failed to read the session user.");
+    return FALSE;
+  }
+  if (getuid() != g_variant_get_uint32(user)) {
+    // the new session isn't for our user
+    return FALSE;
+  }
+  session_type = org_freedesktop_login1_session_get_type_(session);
+  if (session_type == NULL) {
```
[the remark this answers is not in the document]
sergio-costas: But it's not a variant, it's a gchar *...

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,247 @@
...
+        return true;
+      }
+      g_variant_unref(session_object_variant);
+      g_variant_unref(session);
+    }
+  } else {
+    g_message("Failed to get session list (check that login-session-observe "
+              "interface is connected). Forcing a reload.");
+  }
+  return false;
```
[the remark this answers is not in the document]
sergio-costas: ... FALSE? :-) anyway, it's the same reason than before.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,247 @@
...
+    already_in_graphical_session = false;
+    g_main_loop_run(loop);
+  }
+  g_signal_handler_disconnect(login_manager, session_new_id);
+
+  /* Now we know that we are in a graphical session, so now wait for
+   * one of the valid DBus well-known names to be acquired by a process, to
+   * guarantee that the system is ready
+   */
+  g_autoptr(GMainLoop) loop2 = g_main_loop_new(NULL, TRUE);
```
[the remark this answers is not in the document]
sergio-costas: You are right. I wasn't sure if it was possible, but I tested it and seems to work. Done.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,247 @@
...
+ */
+bool sdi_wait_for_graphical_session(void) {
+  bool already_in_graphical_session = true;
+  g_autoptr(GMainLoop) loop = g_main_loop_new(NULL, TRUE);
+  g_autoptr(Login1Manager) login_manager =
+      login1_manager_proxy_new_for_bus_sync(
+          G_BUS_TYPE_SYSTEM, G_DBUS_PROXY_FLAGS_NONE, "org.freedesktop.login1",
+          "/org/freedesktop/login1", NULL, NULL);
+  guint session_new_id = g_signal_connect(login_manager, "session-new",
+                                          G_CALLBACK(new_session), loop);
```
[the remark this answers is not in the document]
sergio-costas: Oh, there is already a comment about it...

snap/snapcraft.yaml
```diff
@@ -23,19 +23,21 @@ plugs:
 apps:
   snapd-desktop-integration:
     extensions: [gnome]
-    command: usr/bin/snapd-desktop-integration
+    command: usr/bin/user-session-helper $SNAP/usr/bin/snapd-desktop-integration
```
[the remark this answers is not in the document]
sergio-costas: Not sure... what happens if one of the command-chain programs/scripts fail? Does it return a failure instead of launching the main program? Because that is what we need...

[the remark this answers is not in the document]
sergio-costas: I will ask in the channel...

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,247 @@
...
+static guint dbus_check_id = 0;
+static guint dbus_timeout_id = 0;
+static gchar *dbus_names[] = {"org.freedesktop.portal.Desktop",
+                              "*org.freedesktop.impl.portal.desktop.", NULL};
+
+static gboolean sdi_session_is_desktop(const gchar *object_path) {
+  g_autoptr(OrgFreedesktopLogin1Session) session = NULL;
+  g_autoptr(GVariant) user = NULL;
+  // these values belongs to the session proxy, so they must not be freed
+  GVariant *user_data = NULL;
```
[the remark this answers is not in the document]
sergio-costas: This is what documentation says:
```
@org.freedesktop.DBus.Property.EmitsChangedSignal("const")
      readonly (uo) User = ...;
```
Anyway, being a GVariant, it makes sense it to be refcounted...

snap/snapcraft.yaml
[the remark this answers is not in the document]
sergio-costas: Tested. It does work.

aleasto (commented):
Logic was hard to follow on a first read, but it seems sound.

Couldn't verify yet because of network issues in the certification lab (which is where I could reproduce the bug)

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,248 @@
...
+  if (check_dbus_names_cb(check_data) != G_SOURCE_REMOVE) {
+    // Wait up to ten seconds before cancelling the wait. At that moment, the
+    // desktop should be launched for sure.
+    dbus_timeout_id =
+        g_timeout_add_once(10000, (GSourceOnceFunc)dbus_timeout_cb, loop);
+    g_main_loop_run(loop);
+    if (dbus_timeout_id != 0) {
+      g_source_remove(dbus_timeout_id);
+      // if the timeout goes off, presume that the DBus names are correct
+      already_in_graphical_session = false;
```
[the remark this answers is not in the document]
sergio-costas: Without xdg-desktop-portal, snaps doesn't work, so it must be installed. About expecting it to be autostarted during desktop startup... well, that's why I added a timeout: for any case where it isn't being autostarted during desktop startup. It happens with Gnome Shell/Ubuntu Desktop, that's why it serves as a provisional hack (which, as everything provisional, will last forever 🤣), so...

But putting a longer timeout might be something sensible...

src/user-session-helper/sdi-user-session-helper.c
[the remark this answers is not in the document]
sergio-costas: Also, it specifically unrefs the value before returning it...
imagen

src/user-session-helper/sdi-user-session-helper.c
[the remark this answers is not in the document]
sergio-costas: Mmm... but then, that means that ANY gtk application will keep the xdg-desktop-portal service restarting over and over, until systemd gives up...

Because what causes to startup xdg-desktop-portal is the call to `gtk_init()`...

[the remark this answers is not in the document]
aleasto: I'm not sure I understand that comment. The shell wouldn't load until graphical-session.target is active, so you don't get to launch gtk applications before that.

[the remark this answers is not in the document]
sergio-costas: Yes, the comment is a little bit misleading, because it is placed in an `if` that is triggered only if the timeout hasn't went off. I'll rewrite that.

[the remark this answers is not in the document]
aleasto: Thanks! I understand now :)

aleasto (commented):
I've verified that the changes fix bug 2042301 for me

aleasto (approved):
Verified again at 568326d

aleasto (approved):

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,255 @@
...
+  }
+  user = g_variant_get_child_value(user_data, 0);
+  if (user == NULL) {
+    g_warning("Failed to read the session user.");
+    return FALSE;
+  }
+  if (getuid() != g_variant_get_uint32(user)) {
+    // the new session isn't for our user
+    return FALSE;
+  }
```
3v1n0, line 75: Why not just using `g_variant_get`? We can avoid quite a bit of code

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,247 @@
...
+    // desktop should be launched for sure.
+    dbus_timeout_id =
+        g_timeout_add_once(10000, (GSourceOnceFunc)dbus_timeout_cb, loop2);
+    g_main_loop_run(loop2);
+    if (dbus_timeout_id != 0) {
+      g_source_remove(dbus_timeout_id);
+      // if the timeout goes off, presume that the DBus names are correct
+      already_in_graphical_session = false;
+    }
+  }
```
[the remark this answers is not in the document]
aleasto: If we have no evidence that snapd is going to fix this by adding some way to define a graphical daemon, then yeah I guess this is better abstracted into a wrapper we can reuse elsewhere.

I also want to ship this fix to users so I'm totally OK with deferring that to later.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,255 @@
...
+    g_autoptr(GVariant) session_object_variant =
+        g_variant_get_child_value(session, 4);
+    const gchar *session_object =
+        g_variant_get_string(session_object_variant, NULL);
+    g_debug("Checking session %s...", session_object);
+    if (sdi_session_is_desktop(session_object)) {
+      g_debug("We are already in a desktop session!");
+      return true;
+    }
+  }
```
3v1n0, line 121: Also here, can just use `g_variant_get` and then iterate over a GVariantIter instead of manual (and partially unchecked) unpacking.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,247 @@
...
+
+static void dbus_timeout_cb(GMainLoop *loop) {
+  dbus_timeout_id = 0;
+  g_main_loop_quit(loop);
+}
+
+static gboolean check_dbus_names_cb(GMainLoop *loop) {
+  g_auto(GStrv) names = NULL;
+  GError *error = NULL;
+  dbus_server__call_list_names_sync(dbus_server, &names, NULL, &error);
```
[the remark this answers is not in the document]
3v1n0: Well, it's just a `g_variant_get`, that will give you the strv anyways...

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,255 @@
...
+    return G_SOURCE_CONTINUE;
+  }
+  for (gchar **dbus_name = dbus_names; *dbus_name != NULL; dbus_name++) {
+    for (gchar **name = names; *name != NULL; name++) {
+      if (*dbus_name[0] == '*') {
+        if (!g_str_has_prefix(*name, (*dbus_name) + 1)) {
+          continue;
+        }
+      } else {
+        if (0 != g_strcmp0(*name, *dbus_name)) {
```
3v1n0, line 158:
```suggestion
        if (g_strcmp0(*name, *dbus_name) != 0) {
```

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,255 @@
...
+   * session IS a graphical one. For that cases we do check if there is a
+   * graphical session active, and if that's the case, we must exit to let
+   * systemd relaunch us again, this time being able to get access to the
+   * session.
+   */
+  if (!sdi_check_graphical_sessions(login_manager)) {
+    already_in_graphical_session = false;
+    g_main_loop_run(loop);
+  }
+  g_signal_handler_disconnect(login_manager, session_new_id);
```
3v1n0, line 203:
```suggestion
  g_clear_signal_handler(&session_new_id, login_manager);
```

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,248 @@
...
+  GError *error = NULL;
+  dbus_server__call_list_names_sync(dbus_server, &names, NULL, &error);
+  if (error != NULL) {
+    g_warning("Failed to get DBus names: %s", error->message);
+    return G_SOURCE_CONTINUE;
+  }
+  for (gchar **dbus_name = dbus_names; *dbus_name != NULL; dbus_name++) {
+    for (gchar **name = names; *name != NULL; name++) {
+      if (*dbus_name[0] == '*') {
+        if (!g_str_has_prefix(*name, (*dbus_name) + 1)) {
```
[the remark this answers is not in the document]
3v1n0: However, can't we use `g_pattern_match_string` here?

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,255 @@
...
+  g_auto(GStrv) names = NULL;
+  GMainLoop *loop = *data;
+  dbusserver *dbus_server = *(data + 1);
+  GError *error = NULL;
+  dbus_server__call_list_names_sync(dbus_server, &names, NULL, &error);
+  if (error != NULL) {
+    g_warning("Failed to get DBus names: %s", error->message);
+    return G_SOURCE_CONTINUE;
+  }
+  for (gchar **dbus_name = dbus_names; *dbus_name != NULL; dbus_name++) {
```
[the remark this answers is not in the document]
sergio-costas: Done.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,255 @@
...
+   * guarantee that the system is ready
+   */
+
+  g_autoptr(dbusserver) dbus_server = dbus_server__proxy_new_for_bus_sync(
+      G_BUS_TYPE_SESSION, G_DBUS_PROXY_FLAGS_NONE, "org.freedesktop.DBus",
+      "/org/freedesktop/DBus", NULL, NULL);
+
+  // we can use a stack-stored array because we use g_main_loop_run(),
+  // so this function will remain alive until the array is not needed anymore.
+  gpointer cb_data[2] = {loop, dbus_server};
```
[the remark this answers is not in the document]
sergio-costas: Done.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,255 @@
...
+ */
+
+#define _GNU_SOURCE
+
+#include "sdi-user-session-helper.h"
+#include <unistd.h>
+
+static guint dbus_check_id = 0;
+static guint dbus_timeout_id = 0;
+static gchar *dbus_names[] = {"org.freedesktop.portal.Desktop",
```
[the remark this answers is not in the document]
3v1n0: Yeah, but it just complicates the code for no reason IMHO... I think we can just rely on simple name watching instead in few lines of code that is already tested and part of Gio:

```c
#include <gio/gio.h>
#include <stdlib.h>

#define NAME_TO_WATCH "org.freedesktop.portal.Desktop"

const int MAX_WAIT = 10; /* In seconds */
static int exit_status = EXIT_FAILURE;

static void name_appeared(GDBusConnection *connection,
                          const gchar *name,
                          const gchar *name_owner,
                          gpointer user_data) {

  GMainLoop *loop = user_data;

  g_debug("Name '%s' appeared (owner: %s)", name, name_owner);
  exit_status = EXIT_SUCCESS;
  g_main_loop_quit(loop);
}

int main(int argc, char *argv[]) {
  g_autoptr(GMainLoop) loop = NULL;
  g_autoptr(GError) error = NULL;
  guint watcher_id;

  loop = g_main_loop_new(NULL, FALSE);

  g_timeout_add_once(MAX_WAIT * 1000, (GSourceOnceFunc) g_main_loop_quit, loop);

  watcher_id = g_bus_watch_name(G_BUS_TYPE_SESSION,
                                NAME_TO_WATCH,
                                G_BUS_NAME_WATCHER_FLAGS_NONE,
                                name_appeared,
                                NULL,
                                loop,
                                NULL);

  g_debug("Waiting for D-Bus name '%s'...", NAME_TO_WATCH);
  g_main_loop_run(loop);

  g_debug("Name " NAME_TO_WATCH " appeared, cleaning up");
  g_bus_unwatch_name(watcher_id);

  return exit_status;
}
```

This works fine when tested for me in a snap, so it seems enough to me.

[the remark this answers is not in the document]
sergio-costas: It seems to work, yes. But the point is that I can't find where in snapd is granted access to subscribe to the `NameOwnerChanged` signal in the `org.freedesktop.DBus` interface. It's like it's working by pure chance.

Of course, I can modify the code and check only for `org.freedesktop.portal.Desktop` to simplify it, but I would stick to the DBus elements that we really know that are allowed in the interfaces we have.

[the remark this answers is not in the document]
sergio-costas: @3v1n0 So it seems that it's not guaranteed that we can connect to `NameOwnerChanged`. Do you want me to modify the code to only check for `org.freedesktop.portal.Desktop`? Or do we just merge it as-is?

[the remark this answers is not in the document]
3v1n0: Well at the moment we can do it I think... I still would prefer to avoid any Impl check, because these are not allowed to access by any snap.

However I'd try to go simpler first, let's just use the way that is normally working (and it does work for all the snaps at the moment), by waiting the name to appear. If we regress in future we can always go back to this, but by that time it's probably likely that the user services won't be experimental anymore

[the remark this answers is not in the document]
sergio-costas: ListNames is permitted in the _system_observer_ and _unity7_ interfaces. Maybe it should be removed from the later, but not from the former one.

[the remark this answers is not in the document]
sergio-costas: In fact, that is added by a lot of interfaces, so...

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,206 @@
...
+  /* check if there is already a graphical session opened for us, in which
+   * case we must just exit and let systemd to relaunch us, because it means
+   * that we run too early and the desktop wasn't still ready.
+   */
+  for (int i = 0; i < g_variant_n_children(sessions); i++) {
+    g_autoptr(GVariant) session = g_variant_get_child_value(sessions, i);
+    if (session == NULL) {
+      continue;
+    }
+    g_autofree gchar *session_object;
```
3v1n0, line 107: It potentially will, but better always to init to NULL these.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,206 @@
...
+      g_str_equal("wayland", session_type) ||
+      g_str_equal("mir", session_type)) {
+    // this is a graphical session
+    return TRUE;
+  }
+  return FALSE;
+}
+
+static gboolean sdi_check_graphical_sessions(Login1Manager *login_manager) {
+  GVariant *sessions = NULL;
```
3v1n0, line 86: Leaked?

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,206 @@
...
+     * session or in a text one, so we will assume that we are in a session
+     * desktop to force a reload.
+     */
+    return TRUE;
+  }
+  g_variant_get(user_data, "(uo)", &user, NULL);
+  if (getuid() != user) {
+    // the new session isn't for our user
+    return FALSE;
+  }
```
3v1n0, line 70: This is not enough though, since we should also check that the seat matches, or we may stop on another seat of the same user. They are not supported yet for users, but we may (there's some work ongoing in gdm).

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,206 @@
...
+  if (session_type == NULL) {
+    g_warning("Failed to read the session type");
+    return FALSE;
+  }
+  if (g_str_equal("x11", session_type) ||
+      g_str_equal("wayland", session_type) ||
+      g_str_equal("mir", session_type)) {
+    // this is a graphical session
+    return TRUE;
+  }
```
3v1n0, line 81: How different is this to instead just check for `$XDG_SESSION_TYPE` or basically? `systemctl --user show-environment |grep XDG_SESSION_TYPE` (as I assume the value may be set later?)?

3v1n0 (commented):
The bus name check is fine, but IMHO we can just simplify the logic avoiding the logind checks or just rely on `$XDG_SESSION_TYPE` or `loginctl show-session self -PType`.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,206 @@
...
+   * a must because, sometimes, snapd-desktop-integration is launched "too
+   * quickly" and the desktop isn't ready, so gtk_init_check() fails but the
+   * session IS a graphical one. For that cases we do check if there is a
+   * graphical session active, and if that's the case, we must exit to let
+   * systemd relaunch us again, this time being able to get access to the
+   * session.
+   */
+  if (!sdi_check_graphical_sessions(login_manager)) {
+    already_in_graphical_session = false;
+    g_main_loop_run(loop);
```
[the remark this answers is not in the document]
sergio-costas: Yes, it ends, because at line 155 we are waiting for a new session opened. The point is to wait as long as needed first for a new session, and only then check if we are already in graphical mode. Don't forget that the idea is to avoid the tons of error messages when an user connects through a SSH session, for example.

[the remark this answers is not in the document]
sergio-costas: If we move that timeout to where you suggest, if the user logs in by SSH, it will time out, launch the daemon, it will fail, an error will be added to the journal, it will be re-launched, will wait again until it times out again, another error... Avoiding that was the reason to add all this mechanism.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: Not sure... I'll test it.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: The problem is that environment variables can be set after this has been launched, and they won't be updated until the code exits and relaunches again. So it's a must to check it here.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: Checked with valgrind after adding a g_autoptr. It doesn't fail, so yes, it's a leak. Fixing...

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: At this moment I think that this is enough, since anyway, the daemon is launched only once.

src/user-session-helper/sdi-user-session-helper.c
3v1n0: > they open a graphical session, the new session will have XDG_SESSION_TYPE, yes, but the helper will still be running in a process with the old set of environment variables

True, but the one launched from SSH should always be stopped anyways, since you don't want it to be running in the graphical session anyways, I assume you can also use `XDG_SESSION_ID` together with logind APIs to check things are you expect.

But the point is that:
 - If the daemon is launched form ssh and another graphical session is launched concurrently, also the current solution is not the best imho, because it implies that the the session triggered by ssh may end up the one _surviving_...
 - How realistic is that a process that just starts and checks the environment, exiting early, can be affected by this?

However I feel that we are putting too much effort for a workaround here, while we could have addressed this in snapd already

sergio-costas: If the daemon is launched from SSH, and then a graphical session is launched, the daemon detects this and re-launches, this time getting access to the graphical session. If the graphical session is killed, the daemon dies with it because GTK loses access to the compositor, and when relaunched will detect only the SSH session and wait again.

The second one... I don't understand that... If you ask how probable is that launching this as we are already doing can cause problems... well, it is ALREADY causing problems: if snapd-desktop-integration is launched too soon, it also launches too soon xdg-desktop-portal and that causes it to don't export some interfaces, like the one for sharing the screen. That's why I'm adding it.

about address this in snapd, yes, I would love to have this fixed in snapd, but that requires either access to the ActiveState property in systemd (which can be done quickly as a new interface, but is still hacky), or try again with SD230 specification.

sergio-costas:
I changed the restart option to "always" because now we are fully sure when we really can launch it, so it's a good idea to use it since it guarantees that any failure in snapd-desktop-integration won't make it "disappear", and it will be relaunched.

3v1n0 (approved):
Thanks, lovely! :)

aleasto:
While the snapd changes are still pending, can we either land this, or at least just ship something silly like sleeping 10 seconds before calling `gtk_init_check` to avoid the race condition where we try to start xdg-desktop-portal before the desktop?

sergio-costas:
Let me rebase this, first, and also check if there is currently a pending candidate...

sergio-costas:
@aleasto There is another problem: there is still an error message going over and over in the journal. I suspect that it's because of GDM... Any ideas?

aleasto:
Can you push the rebase so i can try it?

sergio-costas:
```
abr 24 15:38:00 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227454]: See https://snapcraft.io/docs/home-outside-home for details.
abr 24 15:38:00 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Main process exited, code=exited, status=1/FAILURE
abr 24 15:38:00 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Failed with result 'exit-code'.
abr 24 15:38:01 raster-Laptop-13-AMD-Ryzen-7040Series multipassd[6887]: fetch manifest periodically
abr 24 15:38:02 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Scheduled restart job, restart counter is at 97.
abr 24 15:38:02 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: Started snap.snapd-desktop-integration.snapd-desktop-integration.service - Service for snap application snapd-desktop-integration.snapd-desktop-integration.
abr 24 15:38:02 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227540]: Sorry, home directories outside of /home needs configuration.
abr 24 15:38:02 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227540]: See https://snapcraft.io/docs/home-outside-home for details.
abr 24 15:38:02 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Main process exited, code=exited, status=1/FAILURE
abr 24 15:38:02 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Failed with result 'exit-code'.
abr 24 15:38:04 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Scheduled restart job, restart counter is at 98.
abr 24 15:38:04 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: Started snap.snapd-desktop-integration.snapd-desktop-integration.service - Service for snap application snapd-desktop-integration.snapd-desktop-integration.
abr 24 15:38:05 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227599]: Sorry, home directories outside of /home needs configuration.
abr 24 15:38:05 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227599]: See https://snapcraft.io/docs/home-outside-home for details.
abr 24 15:38:05 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Main process exited, code=exited, status=1/FAILURE
abr 24 15:38:05 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Failed with result 'exit-code'.
abr 24 15:38:07 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Scheduled restart job, restart counter is at 99.
abr 24 15:38:07 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: Started snap.snapd-desktop-integration.snapd-desktop-integration.service - Service for snap application snapd-desktop-integration.snapd-desktop-integration.
abr 24 15:38:07 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227664]: Sorry, home directories outside of /home needs configuration.
abr 24 15:38:07 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227664]: See https://snapcraft.io/docs/home-outside-home for details.
abr 24 15:38:07 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Main process exited, code=exited, status=1/FAILURE
abr 24 15:38:07 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Failed with result 'exit-code'.
abr 24 15:38:09 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Scheduled restart job, restart counter is at 100.
abr 24 15:38:09 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: Started snap.snapd-desktop-integration.snapd-desktop-integration.service - Service for snap application snapd-desktop-integration.snapd-desktop-integration.
abr 24 15:38:09 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227702]: Sorry, home directories outside of /home needs configuration.
abr 24 15:38:09 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227702]: See https://snapcraft.io/docs/home-outside-home for details.
abr 24 15:38:09 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Main process exited, code=exited, status=1/FAILURE
abr 24 15:38:09 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Failed with result 'exit-code'.
abr 24 15:38:11 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Scheduled restart job, restart counter is at 101.
abr 24 15:38:11 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: Started snap.snapd-desktop-integration.snapd-desktop-integration.service - Service for snap application snapd-desktop-integration.snapd-desktop-integration.
abr 24 15:38:11 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227741]: Sorry, home directories outside of /home needs configuration.
abr 24 15:38:11 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227741]: See https://snapcraft.io/docs/home-outside-home for details.
abr 24 15:38:11 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Main process exited, code=exited, status=1/FAILURE
abr 24 15:38:11 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Failed with result 'exit-code'.
abr 24 15:38:13 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Scheduled restart job, restart counter is at 102.
abr 24 15:38:13 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: Started snap.snapd-desktop-integration.snapd-desktop-integration.service - Service for snap application snapd-desktop-integration.snapd-desktop-integration.
abr 24 15:38:14 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227779]: Sorry, home directories outside of /home needs configuration.
abr 24 15:38:14 raster-Laptop-13-AMD-Ryzen-7040Series snapd-desktop-integration.snapd-desktop-integration[1227779]: See https://snapcraft.io/docs/home-outside-home for details.
abr 24 15:38:14 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Main process exited, code=exited, status=1/FAILURE
abr 24 15:38:14 raster-Laptop-13-AMD-Ryzen-7040Series systemd[1222504]: snap.snapd-desktop-integration.snapd-desktop-integration.service: Failed with result 'exit-code'.
```

sergio-costas:
Rebased.

sergio-costas:
Confirmed: it's the GDM session: with no other session opened it still happens, and it specifically complains for the HOME folder :-(

aleasto:
That's not a regression from this change.

aleasto:
It might be a "regression" that s-d-i tries to re-launch itself more often than before (we've changed the restart conditions here)

aleasto:
I think we can ignore it. snapd is broken, but we don't need s-d-i running in the log-in screen. It should stop trying to re-start itself after a few times and not log-spam forever.

sergio-costas:
The problem is that it does spam forever :-(

And the old code doesn't...

sergio-costas:
The restart counter in my log was up to 142, with a 2-seconds respawn...

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,206 @@
...
+}
+
+int main(int argc, char **argv) {
+  if (argc < 2) {
+    g_error("At least one parameter must be passed. Aborting.");
+    return EXIT_FAILURE;
+  }
+  if (!sdi_wait_for_graphical_session()) {
+    g_debug("New graphical session opened. Forcing reload.");
+    return EXIT_FAILURE;
```
aleasto: We can set `restart-condition: on-abnormal` and:

```suggestion
    raise(SIGKILL);
```

aleasto (changes requested):
I think we overlooked a major detail here. XDG Desktop Portal is not automatically started with the desktop. Instead, it is started on-demand whenever an application first tries to call the portal. So watching this name is not a good indication of desktop startup.

src/user-session-helper/sdi-user-session-helper.c
aleasto: No, wait

aleasto: Yeah that should work. According to your journal logs, the snapd failure causes a clean exit rather than an unclean signal which would trigger `on-abnormal`. Can you check if my suggestion works?

aleasto:
> I think we overlooked a major detail here. XDG Desktop Portal is not automatically started with the desktop. Instead, it is started on-demand whenever an application first tries to call the portal. So watching this name is not a good indication of desktop startup.

Here's a new suggestion: watch for the `org.freedesktop.Notifications` name instead.
That's reasonably cross-desktop, and also safe to assume that it is generally started together with the desktop.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: But we must also reload it if the process just ends...

aleasto: In which case does the daemon intentionally exit cleanly?

If it crashes, we already catch it with `on-abnormal`.

sergio-costas: And if the user just kills it...?

sergio-costas:
> > I think we overlooked a major detail here. XDG Desktop Portal is not automatically started with the desktop. Instead, it is started on-demand whenever an application first tries to call the portal. So watching this name is not a good indication of desktop startup.

That's not a problem: the important thing is that the name does appear, not if the process is launched or not.

aleasto:
> > > I think we overlooked a major detail here. XDG Desktop Portal is not automatically started with the desktop. Instead, it is started on-demand whenever an application first tries to call the portal. So watching this name is not a good indication of desktop startup.
> 
> That's not a problem: the important thing is that the name does appear, not if the process is launched or not.

The name only appears once the process has registered on the bus

aleasto:
In practice we have DING and other GTK processes start immediately with the desktop, so xdg-desktop-portal registers immediately. But in general it's not a safe assumption, especially not cross-desktop

src/user-session-helper/sdi-user-session-helper.c
aleasto: If they kill with `SIGKILL`, then we catch it.
If they kill with SIGHUP or SIGTERM, which I believe are caught by the mainloop, we can add another `raise(SIGKILL)` at the end of the `main()`.
Note that if we're currently returning 0 from `main()`, we are already not triggering `on-failure` in this MR

Also, this sounds like the user's problem to be honest...

sergio-costas:
I did a test right now disabling all the extensions and startup programs, and the xdg-desktop-portal process is still being launched.

aleasto:
> I did a test right now disabling all the extensions and startup programs, and the xdg-desktop-portal process is still being launched.

Still, you have something starting the portal, otherwise it wouldn't be starting :)

aleasto:
It might be the shell itself, I don't really know. Point is, it's not as universal as `org.freedesktop.Notifications`.
But we could watch either name if you really want to.

sergio-costas:
But I think that the desktop must launch the main portal to ensure that right services are available... Also, remember that the problem here was that snapd-desktop-integration was activating the portal before everything was ready. Can we guarantee that `org.freedesktop.Notifications` only appears after the desktop has everything ready for the portals?

aleasto:
> But I think that the desktop must launch the main portal to ensure that right services are available

No, generally clients connect on-demand to the portal which connects to the desktop shell, not the other way around

> Can we guarantee that org.freedesktop.Notifications only appears after the desktop has everything ready for the portals

There is no strong guarantee, but I would be comfortable with assuming that.
There is no strong guarantee that xdg-desktop-portal only starts after the desktop is ready either (if there was, we would have no bug at all here...). As of 26.04 we cannot really enforce it: https://salsa.debian.org/debian/xdg-desktop-portal/-/blob/3f27c14e7c2f7be9a8bb63a362650d0e9e651ff5/debian/patches/ubuntu/allow-no-graphical-session-target.patch

aleasto:
Also, again I wouldn't be opposed to add some sneaky `sleep()` calls to make it even more likely :smile:
We're already in workaround territory so I really don't mind...

sergio-costas:
But we are already waiting 10 seconds for the name to appear, and if it doesn't appear, we continue anyway... Check the line 177.

aleasto:
Right.
We can sleep a few seconds *after* detecting the name, before re-launching.
We don't have any constraints to launch immediately...

aleasto:
Ah you mean that we can also rely on the fallback timeout if x-d-p wasn't started? Ok, that's also acceptable. Sorry, it's been a while since we came up with this 😬

Just need to fix the restart condition then.

sergio-costas:
Maybe it's enough to ensure that the session isn't GDM when checking the session type before...

sergio-costas:
Or even simpler: if the HOME folder isn't at `/home`, avoid it!

sergio-costas:
Nah... realized that we can't do the check because it is run after snapd complains...

sergio-costas:
@aleasto Ok, did some changes and now it isn't reloaded when launched in the GDM session, as we want. It should be ready for final revision.

aleasto:
does stopping the systemd service also kill the child process (through cgroup or some other magic...)?

sergio-costas:
Let me test it...

sergio-costas:
Confirmed: it stops and starts as expected.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,232 @@
...
+  memset (&action, 0, sizeof(struct sigaction));
+  action.sa_handler = exit_by_signal;
+  sigaction (signum, &action, NULL);
+}
+
+int main(int argc, char **argv) {
+  if (argc < 2) {
+    g_error("At least one parameter must be passed. Aborting.");
+    // return 1 to avoid reloading it.
+    return EXIT_FAILURE;
```
aleasto: since `EXIT_FAILURE`/`EXIT_SUCCESS` have now lost their meaning, can we switch to explicit numbers?

aleasto (approved):
Thanks. I absolutely hate it, but LGTM.

One nit in-line.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: Done.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: Killing with -9 will return 137. I added code to capture SIGHUP, SIGTERM and SIGINT and return 2 instead of 1, thus ensuring that it reload.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,235 @@
...
+  g_autoptr(OrgFreedesktopLogin1Session) session = NULL;
+  guint32 user;
+  // these values belongs to the session proxy, so they must not be freed
+  GVariant *user_data = NULL;
+  const gchar *session_type = NULL;
+
+  session = org_freedesktop_login1_session_proxy_new_for_bus_sync(
+      G_BUS_TYPE_SYSTEM, G_DBUS_PROXY_FLAGS_NONE, "org.freedesktop.login1",
+      object_path, NULL, NULL);
+  user_data = org_freedesktop_login1_session_get_user(session);
```
3v1n0, line 62: So this is basically using `g_dbus_proxy_get_cached_property` (transfer full), thus use an autoptr to handle `user_data`.

src/user-session-helper/sdi-user-session-helper.c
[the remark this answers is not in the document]
3v1n0: Sure, but a variant string cannot be NULL, unless it's `ms`, so, it can be empty, but never NULL.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,235 @@
...
+   * case we must just exit and let systemd to relaunch us, because it means
+   * that we run too early and the desktop wasn't still ready.
+   */
+  for (int i = 0; i < g_variant_n_children(sessions); i++) {
+    g_autoptr(GVariant) session = g_variant_get_child_value(sessions, i);
+    if (session == NULL) {
+      continue;
+    }
+    g_autofree gchar *session_object = NULL;
+    g_variant_get(session, "(susso)", NULL, NULL, NULL, NULL, &session_object);
```
3v1n0:
```suggestion
    g_variant_get(session, "(&su&s&s&o)", NULL, NULL, NULL, NULL, &session_object);
```

Since you're not using those, I'm not sure if setting NULL, might still lead to some allocation, but in general, we can drop `g_autofree` for the session object here.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,235 @@
...
+int main(int argc, char **argv) {
+  if (argc < 2) {
+    g_error("At least one parameter must be passed. Aborting.");
+    // We are managing '1' as SUCCESS at systemd level, so returning
+    // this will avoid relaunching the daemon.
+    return 1;
+  }
+  set_signal (SIGINT);
+  set_signal (SIGTERM);
+  set_signal (SIGHUP);
```
3v1n0, line 221: Please use `g_unix_signal_add` here.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,235 @@
...
+  // so we must capture it and return a different one to ensure that it is reloaded.
+  // Error code 1 must be only used when we don't want to reload (for example, because
+  // snapd found a problem, like when launching it in GDM).
+  childpid = fork();
+  if (childpid == 0) {
+    execve(argv[1], argv + 1, environ);
+    return 0; // just to avoid warnings in compiler
+  }
+  waitpid(childpid, NULL, 0);
+  return 2; // to ensure that it is reloaded.
```
3v1n0: Use a `#define` so that it's named explaining that is a special return code.

src/user-session-helper/sdi-user-session-helper.c
```diff
@@ -0,0 +1,235 @@
...
+  // when Wayland connection is lost, the application returns with 1 as error code,
+  // so we must capture it and return a different one to ensure that it is reloaded.
+  // Error code 1 must be only used when we don't want to reload (for example, because
+  // snapd found a problem, like when launching it in GDM).
+  childpid = fork();
+  if (childpid == 0) {
+    execve(argv[1], argv + 1, environ);
+    return 0; // just to avoid warnings in compiler
+  }
+  waitpid(childpid, NULL, 0);
```
3v1n0: Why not using `g_spawn*` with `g_child_watch_source_new` to watch?

3v1n0: In any case, maybe in case of failure because of `WIFSIGNALED` or `WIFEXITED != 0` we should return the same?

3v1n0 (commented):
> I absolutely hate it

I agree 😂, not @sergio-costas fault though I hope we can get the snapd folks to improve the situation for these services.

In any case, since the helper is likely re-usable, we should probably include the same into the prompting client for now too.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: Already tried it, but during waitpid it doesn't work because we are outside the main loop. I could use `g_child_watch_source_new()` instead, but I think that it makes the code unnecessary complex (remember that this is just a hack).

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: Good point. Done.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: g_spawn* and g_child_watch... is more complex, and this being a hack I think that it is "good enough".

sergio-costas: About the exit value, the idea is to ensure that, in any failure, systemd relaunches it, always.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: Done.

src/user-session-helper/sdi-user-session-helper.c
sergio-costas: Done.

sergio-costas:
Everything rebased and squashed.

src/user-session-helper/sdi-user-session-helper.c
3v1n0: Yeah, but I would prefer to replicate the child exit status so that we can better catch actual errors of that one, rather than just exit with 2.

What I mean is that we should check the child exit status, and if that's an error we exit in the same way, or we signal the same signal if that's the case.

sergio-costas: The point is that if the child exits with 0 or 1, the daemon will not be relaunched. The only idea that I see is to printf a message with the error value to make it appear in the `systemctl status` log.

aleasto: we can still propagate exit codes other than 0/1 and signals

sergio-costas: We still would be losing info: 0 and 1 are the most common return values.

aleasto: I think the signal codes would be the most interesting part, in case of crash.

sergio-costas: Yes, but since we are relaunching the service again, `systemctl status` won't show the return value anyway. I think that sending it to the log is the best course of action.
image

sergio-costas: Patch sent.

aleasto: Well, it will be logged to the journal by systemd itself as exit status

sergio-costas: Yes, but how can then tell the difference between an error from the child, and an error from the parent code?

aleasto: Fair.
Whatever, this is fine. Please let's just ship something 😭

sergio-costas:
@aleasto @3v1n0 Merged!!!!

PR #150, opened 2025-09-19
Changed the notify_init application name

sergio-costas:
The notify_init call had an incorrect value, which showed an odd title in the notifications.

This patch fixes this.

### DIFF
--- a/src/sdi-notify.c
+++ b/src/sdi-notify.c
@@ -572,7 +572,7 @@ static void sdi_notify_dispose(GObject *object) {
 
 void sdi_notify_init(SdiNotify *self) {
 #ifndef USE_GNOTIFY
-  notify_init("snapd-desktop-integration-test_snapd-desktop-integration-test");
+  notify_init("Snapd Desktop Integration");
 #endif
 }
 


d-loose (approved):

sergio-costas:
Ok, reverted to snapd-desktop-integration_snapd-desktop-integration, but the icon still doesn't appear.

sergio-costas:
@3v1n0 In fact... which patch is the one you mean?

3v1n0:
> Ok, reverted to snapd-desktop-integration_snapd-desktop-integration, but the icon still doesn't appear.

That's two different things, the icon on the launcher is computed by GNOME Shell using the security label of the snap and there's no way for you to change it. If the shell does not find it, it means there's no desktop file named in that way?

> @3v1n0 In fact... which patch is the one you mean?

Go through them to see the logic, but for example https://git.launchpad.net/~aleasto/update-notifier/commit/?h=update-notification&id=a8782b42dcfb73e324ab6bc76f9f3f88a059d628 and https://git.launchpad.net/~aleasto/update-notifier/commit/?h=update-notification&id=5a417aaa5f5583f6f8406bc2ec8d0be5145ff4e7

However (side note on the PR itself), personally I feel that these dialogs are not something we should show by default (for sure not popping out to the user without consent), given we already notify them through a notification and a launcher. But rather we should show them if the user clicks on the notification for more information, and rely just on an indicator meanwhile (I feel we should integrate with the update-notifier indicator that we already have, changing the state when updates are happening in the background).

aleasto:
> I feel that these dialogs are not something we should show by default (for sure not popping out to the user without consent)

This is not too bad, because the dialog only pops up when the user closes an application after receiving the notification asking to do so, so it's sort-of prompted by user interaction.

sergio-costas:
I tested that GJS command, and it works fine:

```
gjs -c "print(imports.gi.GioUnix.DesktopAppInfo.new('snapd-desktop-integration_snapd-desktop-integration.desktop') ?? 'INVALID')"
[object instance wrapper GIName:Gio.DesktopAppInfo jsobj@0xe08bfe7ccb8 native@0x5ef2451ab710]
```

But still fails.

sergio-costas:
In fact, the .desktop file already was marked as dbus activatable. And the no Exec doesn't work: snapcraft complains and refuses to build the snap.

PR #169, opened 2026-04-30
Fix TIC upload

sergio-costas:
The TICS CI fails because one of the tests isn't being run. This PR adds a check for the coverage in the normal tests, and also uses the same command for the TICS CI.

### DIFF
--- a/.github/workflows/tics.yaml
+++ b/.github/workflows/tics.yaml
@@ -50,11 +50,8 @@ jobs:
           wlheadless-run -c weston -- ./_build/tests/test-snapd-desktop-integration
       - name: Test notices monitor
         run: |
-          ./_build/tests/test-sdi-notices-monitor
-          ./_build/tests/test-sdi-progress-dock
-          ./_build/tests/test-sdi-notify
-          ./_build/tests/test-refresh-monitor
-          wlheadless-run -c weston -- ./_build/tests/test-sdi-progress-window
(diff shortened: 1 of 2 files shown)


.github/workflows/tics.yaml
```diff
@@ -50,11 +50,8 @@ jobs:
           wlheadless-run -c weston -- ./_build/tests/test-snapd-desktop-integration
       - name: Test notices monitor
         run: |
-          ./_build/tests/test-sdi-notices-monitor
-          ./_build/tests/test-sdi-progress-dock
-          ./_build/tests/test-sdi-notify
-          ./_build/tests/test-refresh-monitor
-          wlheadless-run -c weston -- ./_build/tests/test-sdi-progress-window
+          # can't use XVFB because gtk_settings uses Xsettings by default, but we need to use keyfile backend
+          wlheadless-run -c weston -- meson test -v -C _build --print-errorlogs -t 3
```
3v1n0, line 54: Can't we use `meson test` as it should now include the everything already?

sergio-costas: It is using meson test... but it must be run inside a wayland session because the tests are graphical.

sergio-costas: In fact, that's why it was failing before: the tests were hardcoded and one was missing. Using meson test will avoid this problem.

PR #170, opened 2026-07-03
Fix g_atomic_ref_count_dec assertion error

sergio-costas:
When `user-session-helper` is run, an error is shown in the journal:

    g_atomic_ref_count_dec: assertion 'old_value > 0' failed

This is because a GVariant is being defined with g_autoptr and freed, when, in reality, it belongs to a container object, thus being released twice.

This patch fixes it.

### DIFF
--- a/src/user-session-helper/sdi-user-session-helper.c
+++ b/src/user-session-helper/sdi-user-session-helper.c
@@ -53,7 +53,7 @@ static gboolean sdi_session_is_desktop(const gchar *object_path) {
   g_autoptr(OrgFreedesktopLogin1Session) session = NULL;
   guint32 user;
   // these values belongs to the session proxy, so they must not be freed
-  g_autoptr(GVariant) user_data = NULL;
+  GVariant *user_data = NULL;
   const gchar *session_type = NULL;
 
   session = org_freedesktop_login1_session_proxy_new_for_bus_sync(


aleasto (approved):
Right, from the `gdbus-codegen` output:

```
/**
 * org_freedesktop_login1_session_get_user: (skip)
 * @object: A #OrgFreedesktopLogin1Session.
 *
 * Gets the value of the <link linkend="gdbus-property-org-freedesktop-login1-Session.User">"User"</link> D-Bus property.
 *
 * Since this D-Bus property is readable, it is meaningful to use this function on both the client- and service-side.
 *
 * The returned value is only valid until the property changes so on the client-side it is only safe to use this function on the thread where @object was constructed. Use org_freedesktop_login1_session_dup_user() if on another thread.
 *
 * Returns: (transfer none) (nullable): The property value or %NULL if the property is not set. Do not free the returned value, it belongs to @object.
 */
GVariant *
org_freedesktop_login1_session_get_user (OrgFreedesktopLogin1Session *object)
{
  g_return_val_if_fail (IS_ORG_FREEDESKTOP_LOGIN1_SESSION (object), NULL);

  return ORG_FREEDESKTOP_LOGIN1_SESSION_GET_IFACE (object)->get_user (object);
}
```

